PR #14, opened 2020-08-18
Adding test for inhomogeneous BCs

ScottMacLachlan:
This adds a test that fails if the treatment of inhomogeneous boundary conditions is not working.  Note that, at present time, this test fails on Firedrake master, but passes on branch smaclachlan/bc-fix.  See firedrakeproject/firedrake#1806)


wence-:
That Firedrake branch is merged, so I guess this could be merged too.

PR #20, opened 2021-02-04
Update ci.yml to trigger GitHub Actions on pull requests

IvanYashchuk:


.github/workflows/ci.yml
```diff
@@ -1,6 +1,6 @@
 name: CI
 
-on: [push]
+on: [push, pull_request]
```
wence-, line 3: I suspect this wants to be `pull_request_target` (see https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target). That way forks can't change ci.yml and run things without us auditing it.

IvanYashchuk: `pull_request_target` seems to be better, thanks!

IvanYashchuk:
Using `pull_request_target` will not trigger the tests now, so this PR shall be merged without waiting for the status to be reported.

wence- (approved):

PR #22, opened 2021-05-16
Fix a few issues with how BCs are implemented

ScottMacLachlan:
There are a few bugs/inconsistencies in how strong BCs are currently implemented.  This aims to fix these, before we add new functionality to enforce BCs in different ways.

### DIFF
--- a/irksome/getForm.py
+++ b/irksome/getForm.py
@@ -156,19 +156,23 @@ def getForm(F, butch, t, dt, u0, bcs=None):
                 gcur = replace(gfoo, {t: t + c[i] * dt})
                 try:
                     gdat = interpolate(gcur, V)
-                except NotImplementedError:
+                    gmethod = lambda g: gdat.interpolate(g)
+                except:  # noqa: E722
                     gdat = project(gcur, V)
-                gblah.append((gdat, gcur))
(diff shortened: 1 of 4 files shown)


irksome/getForm.py
```diff
@@ -156,19 +156,23 @@ def getForm(F, butch, t, dt, u0, bcs=None):
                 gcur = replace(gfoo, {t: t + c[i] * dt})
                 try:
                     gdat = interpolate(gcur, V)
-                except NotImplementedError:
+                    gmethod = lambda g: gdat.interpolate(g)
+                except:  # noqa: E722
```
ScottMacLachlan, line 160: # noqa here (and below) because `interpolate` doesn't always throw a `NotImplementedError` when it isn't implemented

wence-: Can you provide an example/report a Firedrake bug, this is kind of annoying for you.

rckirby:
This is not a firedrake bug per se, just we want to interpolate when we can and project otherwise.  However, Scott is now finding that not all elements that don’t support interpolation fail in the same way/give the same exception, and he has raised this issue.

rckirby (approved):
LGTM.  Thanks for the clean-up!

PR #24, opened 2021-05-19
Implement DAE-style BCs

ScottMacLachlan:
The existing approach to implementing BCs in Irksome is to treat the boundary data as an ODE, implementing `du/dt(x,t) = g'(t)` for a point, x, where a DirichletBC is strongly enforced.

This PR adds functionality to implement the BCs as if they are terms in a DAE, by setting the BCs for the stage values to enforce that the solution reconstruction corresponding each stage should match the BC at the stage time.  For some schemes (LobattoIIIC and RadauIIA, for example), this results is the solution matching the DirichletBC at each timestep.  The theoretical order for this approach matches the stage order of the scheme, while the old approach is limited to (at best) third order.

The old code is preserved, with a switch added in the call to `TimeStepper` or `GetForm` to select between the two approaches.  In the initial commit, all tests are run with the new approach, except for one that touches the LobattoIIIA class of Butcher Tableaux, for which the DAE approach is ill-defined.  One issue for discussion is how many tests to preserve for both approaches.

Finally, this also attempts to fix an oversight in the old approach, whereby it was assumed that the BCs were satisfied by the initial data.  Now, the difference between the solution data at the prior timestep and the prescribed Dirichlet data enters into the stage BCs.

### DIFF
--- a/demos/lowlevel/demo_lowlevel_inhomogbc.py.rst
+++ b/demos/lowlevel/demo_lowlevel_inhomogbc.py.rst
@@ -76,7 +76,7 @@ boundary conditions at each time step::
           dt.assign(1.0 - float(t))
 
       for (gdat, gcur, gmethod) in bcdata:
-          gmethod(gcur)
+          gmethod(gcur,u)
 
       solver.solve()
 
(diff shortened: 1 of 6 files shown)


ScottMacLachlan (commented):
A couple of small clean-ups done.  Now this LGTM

rckirby (approved):

PR #26, opened 2021-06-19
Some manipulation on forms containing time derivatives

wence-:
I can cull the type annotations if you like, though I found them somewhat useful when writing the code.

There's a little bit of cleanup here, and then the main thing is `extract_terms` which splits a form into two, a bit with (linear) time-derivatives in it, and the remainder (that must contain no time-derivatives.

@rckirby, @ScottMacLachlan, I think this will be useful for the automated Galerkin-in-time.

### DIFF
--- a/irksome/getForm.py
+++ b/irksome/getForm.py
@@ -1,17 +1,20 @@
+from functools import reduce
+from operator import mul
+
 import numpy
-from firedrake import (TestFunction, Function, Constant,
-                       split, DirichletBC, interpolate, project)
+from firedrake import (Constant, DirichletBC, Function, TestFunction,
+                       interpolate, project, split)
 from firedrake.dmhooks import push_parent
 from ufl import diff
 from ufl.algorithms import expand_derivatives
-from ufl.classes import Zero
-from .deriv import TimeDerivative  # , apply_time_derivatives
+from ufl.algorithms.analysis import has_exact_type
+from ufl.algorithms.map_integrands import map_integrand_dags
+from ufl.classes import CoefficientDerivative, Zero
 from ufl.constantvalue import as_ufl
 from ufl.corealg.multifunction import MultiFunction
-from ufl.algorithms.analysis import has_exact_type
-from ufl.classes import CoefficientDerivative
-from ufl.algorithms.map_integrands import map_integrand_dags
 from ufl.log import error
+
+from .deriv import TimeDerivative  # , apply_time_derivatives
 
 
 class MyReplacer(MultiFunction):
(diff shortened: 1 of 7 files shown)


irksome/getForm.py
```diff
@@ -1,18 +1,18 @@
...
-from ufl.corealg.multifunction import MultiFunction
 from ufl.algorithms.analysis import has_exact_type
-from ufl.classes import CoefficientDerivative
 from ufl.algorithms.map_integrands import map_integrand_dags
+from ufl.classes import CoefficientDerivative, Zero
+from ufl.constantvalue import as_ufl
+from ufl.corealg.multifunction import MultiFunction
 from ufl.log import error
 
+from .deriv import TimeDerivative  # , apply_time_derivatives
```
wence-, line 14: I have `py-isort` automatically turned on which sorts imports into a canonical order, I can undo this change if you want.

irksome/getForm.py
```diff
@@ -1,18 +1,18 @@
...
+
 
 class MyReplacer(MultiFunction):
```
wence-, line 17: FWIW, this code is _exactly the same_ as the `replace` functionality in UFL, so could be removed.

irksome/getForm.py
```diff
@@ -127,13 +127,15 @@ def getForm(F, butch, t, dt, u0, bcs=None, bc_type="DAE"):
     V = v.function_space()
     assert V == u0.function_space()
 
-    A = numpy.array([[Constant(aa) for aa in arow] for arow in butch.A])
-    c = numpy.array([Constant(ci) for ci in butch.c])
+    A = numpy.array([[Constant(aa) for aa in arow] for arow in butch.A],
+                    dtype=object)
+    c = numpy.array([Constant(ci) for ci in butch.c],
+                    dtype=object)
```
wence-, line 133: These `dtype=object` annotations remove a numpy warning about instantiating nested numpy arrays with objects.

irksome/stepper.py
```diff
@@ -70,16 +70,10 @@ def _update(self):
...
-                u0.dat.data[:] += dtc * b[i] * ks[i].dat.data[:]
-        else:
-            k = self.stages
-
-            for s in range(ns):
-                for i in range(nf):
-                    u0.dat.data[i][:] += dtc * b[s] * k.dat.data[nf*s+i][:]
+        ks = self.ks
+        for s in range(ns):
+            for i, u0d in enumerate(u0.dat):
```
wence-, line 75: A dat, even if not mixed, is iterable (yielding itself in the non-mixed case) so one doesn't have to special case mixed vs. non-mixed in the stage application. Since the `ks` are also only read, I access with `.data_ro` which will be slightly cheaper.

rckirby: Didn't realize this.  The also simplifies the logic going forward.

irksome/getForm.py
rckirby: Thanks, good catch

irksome/getForm.py
rckirby: No worries, I don't really mind.

irksome/getForm.py
rckirby: Last year, we had discussed this -- I had thought I was doing post-order while UFL was doing pre-order traversal.  I was wanting to make sure `Dt(u)` got replaced by itself without `u` first being handled.

irksome/manipulation.py
```diff
@@ -0,0 +1,194 @@
+"""Manipulation of expressions containing :class:`~.TimeDerivative`
```
rckirby, line 1: Do we want to put in some kind of tests for this now, or save that until we start doing Galerkin-type manipulations down the road?

irksome/getForm.py
wence-: Ah, my bad, missed that.

irksome/manipulation.py
wence-: Added some tests.

rckirby (approved):

PR #28, opened 2021-07-12
Rckirby/redo stages

rckirby:
This PR enables the common reformulation to solve for stage variables `w = Ak` instead of `k`, which makes the mass matrix more dense but makes the stiffness part block diagonal.  This form is used in many contemporary preconditioning papers.

The main technique is introducing an optional argument into `getForm` that factors the Butcher matrix `A` into a pair of matrices `A = A1 A2`.  The classical RK method (default) puts `A1 = A` and `A2 = I` and the reformulation uses the reverse.  The new code should (untested) supports other choices like `A = L U`, although it is unclear whether there would be any mathematical and/or computational advantage to such approaches.  

In many cases for RK methods of collocation type the `A = I A` factorization also leads to a much faster update of the solution (increment the current state by `dt` times the final stage variable, and the `Stepper` constructor detects this case and assigns a member `_update` method accordingly.

### DIFF
--- a/irksome/ButcherTableaux.py
+++ b/irksome/ButcherTableaux.py
@@ -24,10 +24,6 @@ class ButcherTableau(object):
         self.btilde = btilde
         self.c = c
         self.order = order
-        try:
-            self.Ainv = numpy.linalg.inv(A)
-        except numpy.linalg.LinAlgError:
-            self.Ainv = None
 
     @property
     def num_stages(self):
--- a/irksome/getForm.py
+++ b/irksome/getForm.py
@@ -82,7 +82,16 @@ class BCStageData(object):
         self.gstuff = (gdat, gcur, gmethod)
 
 
-def getForm(F, butch, t, dt, u0, bcs=None, bc_type="DAE"):
+# Utility functions that help us refactor
+def AI(A):
+    return (A, numpy.eye(*A.shape, dtype=A.dtype))
+
+
+def IAinv(A):
+    return (numpy.eye(*A.shape, dtype=A.dtype), numpy.linalg.inv(A))
+
+
+def getForm(F, butch, t, dt, u0, bcs=None, bc_type="DAE", splitting=AI):
     """Given a time-dependent variational form and a
     :class:`ButcherTableau`, produce UFL for the s-stage RK method.
 
@@ -93,6 +102,11 @@ def getForm(F, butch, t, dt, u0, bcs=None, bc_type="DAE"):
          Any explicit time-dependence in F is included
     :arg dt: a :class:`Constant` referring to the size of the current
          time step.
+    :arg splitting: a callable that maps the (floating point) Butcher matrix
+         a to a pair of matrices `A1, A2` such that `butch.A = A1 A2`.  This is used
+         to vary between the classical RK formulation and Butcher's reformulation
+         that leads to a denser mass matrix with block-diagonal stiffness.
+         Some choices of function will assume that `butch.A` is invertible.
     :arg u0: a :class:`Function` referring to the state of
          the PDE system at time `t`
     :arg bcs: optionally, a :class:`DirichletBC` object (or iterable thereof)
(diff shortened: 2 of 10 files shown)


irksome/getForm.py
```diff
@@ -124,60 +142,81 @@ def getForm(F, butch, t, dt, u0, bcs=None, bc_type="DAE"):
...
+
+    A1 = numpy.array([[Constant(aa) for aa in arow] for arow in bA1],
+                     dtype=object)
+    if bA1inv is not None:
+        A1inv = numpy.array([[ConstantOrZero(aa) for aa in arow] for arow in bA1inv],
+                            dtype=object)
+    else:
+        A1inv = None
+    A2inv = numpy.array([[ConstantOrZero(aa) for aa in arow] for arow in bA2inv],
+                        dtype=object)
```
wence-, line 172: `bA2inv` may be `None` at this point, so you could get a `TypeError` here.

irksome/getForm.py
```diff
@@ -124,60 +142,81 @@ def getForm(F, butch, t, dt, u0, bcs=None, bc_type="DAE"):
...
+    wbits_np = numpy.zeros((num_stages, num_fields), dtype=object)
 
     for i in range(num_stages):
         for j in range(num_fields):
-            kbits_np[i, j] = kbits[i*num_fields+j]
+            wbits_np[i, j] = wbits[i*num_fields+j]
 
-    Ak = A @ kbits_np
+    A1w = A1 @ wbits_np
+    A2invw = A2inv @ wbits_np
```
wence-, line 207: You always need `A2inv`, but sometimes it isn't available, I suspect you want to error sooner.

irksome/getForm.py
wence-: Since you always need `A2inv` probably you want to say
```
if bA2inv is None:
    raise NotImplementedError("... is not invertible, can't continue")
```

irksome/stepper.py
```diff
@@ -38,7 +39,8 @@ class TimeStepper:
 
     """
     def __init__(self, F, butcher_tableau, t, dt, u0, bcs=None,
-                 solver_parameters=None, bc_type="DAE"):
+                 solver_parameters=None, bc_type="DAE", splitting=AI):
+        self.splitting = splitting
```
wence-, line 43: `splitting` is never needed outside `__init__` so could just not save it.

irksome/stepper.py
```diff
@@ -56,24 +58,46 @@ def __init__(self, F, butcher_tableau, t, dt, u0, bcs=None,
...
 
         if self.num_stages == 1 and self.num_fields == 1:
-            self.ks = (stages,)
+            self.ws = (stages,)
         else:
-            self.ks = stages.split()
+            self.ws = stages.split()
+
+        A1, A2 = splitting(butcher_tableau.A)
+        self.updateb = numpy.linalg.solve(A2.T, butcher_tableau.b)
```
wence-, line 66: A2 may not be invertible so catch `LinAlgError` here and report something useful?

tests/test_accuracy.py
```diff
@@ -53,10 +54,10 @@ def heat(n, deg, time_stages):
...
 
 
-@pytest.mark.parametrize(('deg', 'convrate', 'time_stages'),
-                         [(1, 1.78, i) for i in (1, 2)]
-                         + [(2, 2.8, i) for i in (2, 3)])
-def test_heat_eq(deg, convrate, time_stages):
+@pytest.mark.parametrize(('deg', 'convrate', 'time_stages', 'splitting'),
+                         [(1, 1.78, i, splt) for i in (1, 2) for splt in (AI, IA)]
+                         + [(2, 2.8, i, splt) for i in (2, 3) for splt in (AI, IA)])
+def test_heat_eq(deg, convrate, time_stages, splitting):
```
wence-, line 60: FWIW, a much nicer way of doing this is to just use multiple `parameterize` markers:
```
@pytest.mark.parametrize("splitting", (AI, IA))
@pytest.mark.parametrize("deg", "convrate", "time_stages", ...)
def test_heat_eq(deg, convrate, time_stages, splitting):
   ...
```

wence- (commented):
Some minor comments

irksome/getForm.py
rckirby: I think that we require `A2inv` everywhere.  So you're right, we should just error when we try inverting `bA2`.

irksome/stepper.py
rckirby: Correct,  not sure if we'll ever need it or not, but it's easy to put back if we do.

irksome/getForm.py
rckirby: Yes, it can be done when failing to form `bA2inv`

tests/test_accuracy.py
rckirby: Yes, thanks, fixed.

irksome/stepper.py
rckirby: `getForm` would have failed if `A2` were not invertible (see above discussion).  I could have `getForm` return the factors, but computing them is cheap and the return value from `getForm` is already pretty complicated.  But I added a try/except around this.

irksome/getForm.py
```diff
@@ -124,60 +142,81 @@ def getForm(F, butch, t, dt, u0, bcs=None, bc_type="DAE"):
...
+    except numpy.linalg.LinAlgError:
+        bA1inv = None
+    try:
+        bA2inv = numpy.linalg.inv(bA2)
+        A2inv = numpy.array([[ConstantOrZero(aa) for aa in arow] for arow in bA2inv],
+                            dtype=object)
+    except numpy.linalg.LinAlgError:
+        raise NotImplementedError("We require A = A1 A2 with A2 invertible")
+
+    A1 = numpy.array([[Constant(aa) for aa in arow] for arow in bA1],
```
ScottMacLachlan, line 166: Is there any particular reason to not use `ConstantOrZero(aa)` here?

irksome/stepper.py
```diff
@@ -56,24 +57,49 @@ def __init__(self, F, butcher_tableau, t, dt, u0, bcs=None,
...
+        ws = self.ws
         for s in range(ns):
             for i, u0d in enumerate(u0.dat):
-                u0d.data[:] += dtc * b[s] * ks[nf*s+i].dat.data_ro
+                u0d.data[:] += dtc * b[s] * ws[nf*s+i].dat.data_ro
+
+    def _update_A2Tmb(self):
+        """Assuming the algebraic problem for the RK stages has been
+        solved, updates the solution.  This will not typically be
+        called by an end user."""
```
ScottMacLachlan, line 94: I feel like this needs a slightly more pointed docstring - isn't the point here that this is _just_ the special case for `w = Ak`?

irksome/ButcherTableaux.py
```diff
@@ -24,10 +24,6 @@ def __init__(self, A, b, btilde, c, order):
         self.btilde = btilde
         self.c = c
         self.order = order
-        try:
-            self.Ainv = numpy.linalg.inv(A)
-        except numpy.linalg.LinAlgError:
-            self.Ainv = None
```
ScottMacLachlan, line 27: I think that all of the code that relied on this has been removed in this PR, so we should remove this as well.

irksome/getForm.py
rckirby: Good catch.  When `A = I A` you'd like to catch the sparsity.

irksome/ButcherTableaux.py
rckirby: `grep -iR ainv */*py` suggests that this is the case.  If we need it again in the future it's easy to put back.

irksome/stepper.py
rckirby: Yes, I copied/pasted but didn't update it.

ScottMacLachlan (approved):
Looks good to me!  Thanks!

wence- (approved):

PR #31, opened 2021-12-10
Passing nullspace through to treat DAEs like Stokes

### LINKED ISSUE #23: best way to assign a nullspace?
Hi, wondering what methods I can use to pass nullspace information about the system to the solver? I see some commented lines in the Navier Stokes demos but nothing implemented. Thanks!

ScottMacLachlan:
This adds functionality for treating certain nullspaces in Irksome, aimed at the case of DAEs, where the nullspace is associated with a constraint.

Given the original product space, we take in an indexed list of subspace indices and nullspace bases, then construct the `MixedVectorSpaceBasis` over the Irksome product space, matching components of that space to this list.  I'm not sure this is the most elegant implementation, but it seems to work.  (I'm open to suggestions here, though!)

I've modified the `test_stokes.py` code to include the nullspace for Stokes in this way, and verified that passing `nullspace=None` in the call to `stepper` produces failing tests (pressure errors are large, but constant, as expected), while including the nullspace produces the desired results.

This would seem to resolve #23.

Note to self (because it tripped me up when testing on an old install): this requires https://github.com/firedrakeproject/firedrake/pull/1930 , where the treatment of nontrivial `MixedVectorSpaceBasis` was fixed.

### DIFF
--- a/demos/lowlevel/demo_lowlevel_homogbc.py.rst
+++ b/demos/lowlevel/demo_lowlevel_homogbc.py.rst
@@ -54,7 +54,7 @@ Continuing::
 
 Now, we use the :func:`.getForm` function, which processes the semidiscrete problem::
 
-  Fnew, k, bcnew, bcdata = getForm(F, butcher_tableau, t, dt, u, bcs=bc)
+  Fnew, k, bcnew, nspnew, bcdata = getForm(F, butcher_tableau, t, dt, u, bcs=bc)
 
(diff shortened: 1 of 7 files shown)


rckirby (approved):
I was going to grip that you broke the `RanaPC` since it uses `getForm`.  But I guess the test suites caught that!

LGTM, thanks, and sorry for the delay in reviewing.

PR #36, opened 2022-04-07
Rckirby/fixbc

rckirby:
Fixes an error noticed by @medinaeder on imposing `DirichletBC` on components of a vector-valued space.


ScottMacLachlan (approved):
This looks good to me.

PR #37, opened 2022-04-07
Manipulating time derivative forms

rckirby:
Previously, the facility to split terms with time derivatives from the rest of a form was broken in the case of taking `Dt` of a piece of a mixed function space.  This fixes that and adds a test to check.

### DIFF
--- a/irksome/manipulation.py
+++ b/irksome/manipulation.py
@@ -76,7 +76,7 @@ Result = Union[Tuple[()], Tuple[Coefficient, ...]]
 
 @singledispatch
 def _check_time_terms(o, self: Memoizer) -> Result:
-    raise AssertionError
+    raise AssertionError(f"Unhandled type {type(o)}")
 
 
 @_check_time_terms.register(TimeDerivative)
@@ -153,7 +153,8 @@ def check_integrals(integrals: List[Integral], expect_time_derivative: bool = Tr
     howmany = int(expect_time_derivative)
     if len(time_derivatives - {()}) != howmany:
         raise ValueError(f"Expecting time derivative applied to {howmany}"
-                         f"coeffficients, not {len(time_derivatives - {()})}")
+                         f"coefficients, not {len(time_derivatives - {()})}")
+    return integrals
 
 
 def summands(o: Expr) -> FrozenSet[Expr]:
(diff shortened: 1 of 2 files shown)


irksome/manipulation.py
```diff
@@ -152,8 +153,7 @@ def check_integrals(integrals: List[Integral], expect_time_derivative: bool = Tr
     howmany = int(expect_time_derivative)
     if len(time_derivatives - {()}) != howmany:
         raise ValueError(f"Expecting time derivative applied to {howmany}"
-                         f"coefficients, not {len(time_derivatives - {()})}")
-    return integrals
+                         f"coeffficients, not {len(time_derivatives - {()})}")
```
wence-, line 156: This is a regression from main which you should drop.

irksome/manipulation.py
```diff
@@ -190,6 +190,6 @@ def extract_terms(form: Form) -> SplitTimeForm:
         if not isinstance(rest, Zero):
             rest_terms.append(integral.reconstruct(integrand=rest))
 
-    time_terms = check_integrals(time_terms, expect_time_derivative=True)
-    rest_terms = check_integrals(rest_terms, expect_time_derivative=False)
+    check_integrals(time_terms, expect_time_derivative=True)
+    check_integrals(rest_terms, expect_time_derivative=False)
```
wence-, line 194: This is regression from main which you should drop.

irksome/manipulation.py
```diff
@@ -75,7 +76,7 @@ def remove_if(expr: Expr, predicate: Callable[[Expr], bool]) -> Expr:
 
 @singledispatch
 def _check_time_terms(o, self: Memoizer) -> Result:
-    raise AssertionError(f"Unhandled type {type(o)}")
+    raise AssertionError
```
wence-, line 79: This is a regression from main which you should drop.

irksome/manipulation.py
```diff
@@ -84,7 +85,7 @@ def _check_timederiv(o: TimeDerivative, self: Memoizer) -> Result:
     if self(op):
         # op already has a TimeDerivative applied to it
         raise ValueError("Can only handle first-order systems")
-    terminals = tuple(set(traverse_unique_terminals(op)))
+    terminals = tuple(set([x for x in traverse_unique_terminals(op) if not isinstance(x, MultiIndex)]))
```
wence-, line 88: I think this looks right.

wence- (changes requested):
Somehow this removes a bunch of return values and informational stuff.

ScottMacLachlan (commented):
I have nothing to add to this discussion

wence- (approved):

PR #41, opened 2022-05-18
Rckirby/stage

rckirby:
This rather substantial PR introduces the ability to formulate RK methods in terms of the solution values at the time stages rather than the classical derivative formulation.  Advantages:

- Boundary conditions are very straightforward here
- This is the right formulation to generalize to IMEX-type methods of Buvoli (and that's the next step.

On the other hand, except for RadauIIA methods, one must also solve a single-stage mass matrix to advance the type step (if you are willing to pay for fully implicit RK methods, you have no right to complain about this).

The `TimeStepper` class has been replaced by a Python function that dispatches to an appropriate time-stepper class for either stage values or stage derivatives.  The default is to the stage derivative formulation so that the older behavior is preserved.

Also, in shuffling the architecture around classes for time steppers, I removed `AdaptiveTimeStepper` since it was largely unused and of doubtful correctness (order differences of 1 with embedded methods are hard to come by in the implicit RK world).  It would be useful to put back in -- correctly -- at some point.


irksome/stepper.py
```diff
@@ -1,14 +1,82 @@
...
+            F(t, u; v) == 0, where `u` is the unknown
+            :class:`firedrake.Function and `v` iss the
+            :class:firedrake.TestFunction`.
+    :arg butcher_tableau: A :class:`ButcherTableau` instance giving
+            the Runge-Kutta method to be used for time marching.
+    :arg t: A :class:`firedrake.Constant` instance that always
+            contains the time value at the beginning of a time step
+    :arg dt: A :class:`firedrake.Constant` containing the size of the
+            current time step.  The user may adjust this value between
+            time steps, but see :class:`AdaptiveTimeStepper` for a
```
ScottMacLachlan, line 29: Probably remove comment about `AdaptiveTimeStepper` for now

ScottMacLachlan: Also below in docstring for `StageDerivativeTimeStepper`

irksome/getForm.py
```diff
@@ -3,63 +3,16 @@
 
 import numpy
 from firedrake import (Constant, DirichletBC, Function, TestFunction,
-                       interpolate, project, split, MixedVectorSpaceBasis)
+                       interpolate, project, split)
+# from firedrake.dmhooks import push_parent
```
ScottMacLachlan, line 7: Remove this?

irksome/stage.py
```diff
@@ -0,0 +1,326 @@
...
+        vbits = np.array(list(split(v)), dtype="O")
+        VVbits = np.reshape(split(VV), nsxnf)
+        UUbits = np.reshape(split(UU), nsxnf)
+
+    split_form = extract_terms(F)
+
+    Fnew = Zero()
+
+    # print(split_form.time)
+    # print(split_form.remainder)
```
ScottMacLachlan, line 61: Remove debugging output

irksome/stage.py
```diff
@@ -0,0 +1,326 @@
...
+
+from .tools import replace, getNullspace, AI, IA
+from .manipulation import extract_terms, strip_dt_form
+from ufl.classes import Zero
+from .ButcherTableaux import RadauIIA
+from numpy import vectorize
+
+
+def getFormStage(F, butch, u0, t, dt, bcs=None, splitting=None,
+                 nullspace=None):
```
ScottMacLachlan, line 18: Does this need a docstring?

ScottMacLachlan (commented):
This generally looks good to me, aside from a few small cleanups.  (Disclaimer: I didn't check all the math details, but I trust the tests wouldn't pass if the math was wrong.)

demos/cahnhilliard/demo_cahnhilliard.py.rst
```diff
@@ -134,17 +134,17 @@ Save the initial condition to a file::
 
 Now let's do an adaptive time stepper::
```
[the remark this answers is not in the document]
rckirby: Good catch, there was another "adaptive" reference i also caught.

irksome/stepper.py
rckirby: Yes, good catch.

irksome/stage.py
rckirby: Done.

irksome/stage.py
rckirby: Added.

irksome/getForm.py
rckirby: Yes, done.

ScottMacLachlan (approved):
LGTM

Issue #46, opened 2022-07-04
Nullspace errors not caught properly


ScottMacLachlan:
When we check the formatting of the null space being passed in [here](https://github.com/firedrakeproject/Irksome/blob/0974dbc9484522e2f2bbf89f361cbca6ade9eab7/irksome/tools.py#L31), we check for a `TypeError`, but if you put in a `MixedVectorSpaceBasis`, sort throws an `AttributeError`.  We should fix that...

ScottMacLachlan:
Fix proposed in #47

ScottMacLachlan:
Fixed in #47

PR #47, opened 2022-07-06
Change TypeError to AttributeError

ScottMacLachlan:
This is a fix for #46 , changing the `TypeError` to an `AttributeError`.  My only question is whether there's a reasonable use case where `nullspace.sort()` will throw a `TypeError` and we should be catching both...  Thoughts?

### DIFF
--- a/irksome/tools.py
+++ b/irksome/tools.py
@@ -28,8 +28,8 @@ def getNullspace(V, Vbig, butch, nullspace):
     else:
         try:
             nullspace.sort()
-        except TypeError:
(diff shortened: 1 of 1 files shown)


rckirby:
So does this actually allow the user to pass in a `MixedVectorSpaceBasis` rather than the current list structure? `MVSB` is the more common usage in the rest of Firedrake land.

ScottMacLachlan:
No - this just prints out the existing error message when the user tries to pass in an `MVSB` by mistake.  Right now, that message doesn't get reported, because we don't trip the `except` block correctly.

PR #51, opened 2022-09-09
Rckirby/gently

rckirby:
Referees (and occasionally users) have asked for support for DIRKs.  For linear problems, existing demos DTRT with just using solver parameters (although they use too much memory building a triangular system for all the stages rather than one system that's reused for all stages) but it's hard to steer nonlinear problems to solve for one stage at a time.

### DIFF
--- a/irksome/tools.py
+++ b/irksome/tools.py
@@ -1,5 +1,5 @@
 import numpy
-from firedrake import Constant, MixedVectorSpaceBasis, split
+from firedrake import MixedVectorSpaceBasis, split
 from ufl.algorithms.analysis import extract_type, has_exact_type
 from ufl.algorithms.map_integrands import map_integrand_dags
 from ufl.classes import CoefficientDerivative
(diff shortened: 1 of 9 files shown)


rckirby:
After opening the PR, it seems like a better option for the interface might be to let `TimeStepper` take an optional kwarg `dirk=False` that, if true, returns a `DIRKTimeStepper` rather than requiring the user to invoke the constructor.  It would be more similar to the other options we've put in.

irksome/tools.py
```diff
@@ -9,6 +9,9 @@
 from irksome.deriv import TimeDerivative
 
 
+vecconst = numpy.vectorize(Constant)
```
ScottMacLachlan, line 12: This doesn't seem to be being used, but duplicates a line from `stage.py`.  Am I missing something or is this truly orphaned?

ScottMacLachlan (commented):
I'm committing a couple of small cleanups, but otherwise this looks good to me.

irksome/tools.py
rckirby: At some point I was planning to stick `vecconst` in `tools.py` but this line is currently orphaned.

PR #53, opened 2022-09-21
Use self hosted runners

JDBetteridge:


rckirby:
Is this to fix the problem with the frequent failure of docs to build?

JDBetteridge:
Correct! I meant to do this last week but got side tracked

PR #55, opened 2022-09-26
Rckirby/kwargs

rckirby:
Unify access to various methods (RK with different stage types, dirks, imex) all inside the `TimeStepper` interface, clean up kwarg checking.


ScottMacLachlan (approved):
Sorry for being slow - this all looks good to me!

Issue #52, opened 2022-09-14
Is Irksome adjoinable?


rckirby:
Updating via dats is bad.
We should fix that
Also, @dham has suggested putting in an adjoint Taylor test just to make sure we don't break anything.

rckirby:
There is an example of doing a Taylor test for a hand-rolled time stepper in dolfin-adjoint here:
https://www.dolfin-adjoint.org/en/latest/documentation/verification.html

PR #56, opened 2022-10-19
Remove dats from updates

ScottMacLachlan:
In the original implementation, we updated various things by working directly on the dat's of the functions involved.  That's problematic if you want to do adjoint stuff, since dat updates aren't adjoint-able.  This removes the dat accesses in favour of arithmetic on functions or assign statements, addressing #52 .


rckirby (approved):
Thanks!  LGTM

PR #61, opened 2023-01-27
Wrap up bc args

ScottMacLachlan:
A simple test case using stage type value failed when called with a zero DirichletBC, because we weren't wrapping `bc._original_arg` with `as_ufl`.  This should fix that for all places that we use it.


rckirby (approved):
This seems fine to me.

PR #66, opened 2023-02-08
add domains to constants

rckirby:
Currently, Firedrake can hang in parallel on garbage collection, and a known workaround is to attach meshes to `Constant` objects.


ScottMacLachlan (approved):
LGTM (if the tests pass!)

PR #72, opened 2023-03-21
Rckirby/derogue setup

rckirby:
This removes a bit of code in the installer that @JDBetteridge reports can create problems.
The (minor) downside is that one could install Irksome without Firedrake, which doesn't do much... We also update the README.md to nudge people to install with `firedrake-install`/`firedrake-update`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,4 +4,4 @@ This package works with Firedrake to generate Runge-Kutta methods from a semi-di
 
 A long-standing critique of fully implicit RK methods, especially for PDE, is that they require a very large algebraic solve for all stages concurrently.  However, we can use Firedrake's solver infrastructure to address this issue, and also recover most of the comparative efficiency of DIRK or explicit methods.
 
-The core of Irksome is based on UFL manipulation and so should be adaptable to work with FEniCS or other UFL-based packages, but the current version works only with Firedrake.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -4,4 +4,4 @@ This package works with Firedrake to generate Runge-Kutta methods from a semi-di
 
 A long-standing critique of fully implicit RK methods, especially for PDE, is that they require a very large algebraic solve for all stages concurrently.  However, we can use Firedrake's solver infrastructure to address this issue, and also recover most of the comparative efficiency of DIRK or explicit methods.
 
-The core of Irksome is based on UFL manipulation and so should be adaptable to work with FEniCS or other UFL-based packages, but the current version works only with Firedrake.
+The core of Irksome is based on UFL manipulation and so should be adaptable to work with FEniCS or other UFL-based packages, but the current version works only with Firedrake.  As such, it requires a working Irksome installation.  We recommend installing Irksome via the `--install irksome` option.  Given a preexisting Firedrake installation, one may obtain Irksome with options to `firedrake-update`.
```
ScottMacLachlan, line 7: I think it should be clear that the `--install irksome` option is for the _firedrake_ installer, not within Irksome itself.

ScottMacLachlan, line 7: Do we want to add instructions for installing without running `firedrake-install` or `firedrake-update`?  Something like: "You can also install by running `python setup.py install` inside a clone of this repo within an active firedrake venv."?

rckirby: Good point.

rckirby: I added some text to this effect as well.

ScottMacLachlan (approved):

README.md
ScottMacLachlan: LGTM

PR #73, opened 2023-08-30
Smaclachlan/unconstant

ScottMacLachlan:
Using `Constant` for time and dt is now not supported by firedrake, so this PR introduces a `MeshConstant` in `irksome/tools.py` and uses that instead.

Also fixes a change to install `pylit` via pip instead of within the firedrake src dir.


rckirby (approved):
This looks straightforward but pervasive.  Thanks for handling, and I hope you used a search/replace feature...
(Write sed!)

PR #74, opened 2023-09-07
Missed one hardcoded pylit

ScottMacLachlan:
Like the title says, I missed one instance of a hardcoded `pylit` path in #73 . This fixes it.


rckirby (approved):
Oops, I missed this a while back

PR #77, opened 2023-11-17
Ensure DOFs are ordered by coordinate

### LINKED ISSUE #76: LobattoIIIC tests are broken
Perhaps Gauss-Lobatto nodes got reordered in a recent FIAT PR?

pbrubeck:
After https://github.com/firedrakeproject/fiat/pull/49 GLL DOFs inherit the same ordering as Lagrange, which is no longer by coordinate, but now vertex DOFs are ordered first, followed by the interior DOFs. This PR should fix the failing tests.

fixes #76


rckirby (approved):
Thanks for the fix!

PR #79, opened 2024-01-26
Removing Constant domains everywhere I can find them

ScottMacLachlan:
In #73 , we removed `Constant` for all of our `time` and `dt` objects, by introducing `MeshConstant` in `irksome/tools.py`.  This finishes(?) that job, by removing other places where we assigning constants with domains, and using `MeshConstant` for all of these


rckirby (commented):
I think this is fine.  Do we understand what needs to be a `Constant` versus having an attached mesh?   Is it anything that might get recycled (which a butcher tableau might if you had a loop over stepping methods)?

PR #81, opened 2024-02-27
Rckirby/moredirks

rckirby:
Add dirks with higher weak stage order from Ketcheson's work.


ScottMacLachlan (approved):
Not certifying that I've checked every digit of every DIRK, but this looks good to me.

PR #82, opened 2024-02-27
Shortening some tests/demos

rckirby:
Tweak tests to take less time to run.


tests/test_stokes.py
```diff
@@ -45,22 +49,42 @@ def StokesTest(N, butcher_tableau, stage_type="deriv", splitting=AI):
...
+    p.interpolate(pexact-assemble(pexact*dx))
+
+    ns = butcher_tableau.num_stages
+    ind_pressure = ",".join([str(2*i+1) for i in range(ns)])
+    solver_params = {
+        "mat_type": "aij",
+        "snes_type": "ksponly",
+        "ksp_type": "fgmres",
+        "ksp_max_it": 200,
+        "ksp_gmres_restart": 30,
```
ScottMacLachlan, line 61: Not critical for a test, but these seem like generic defaults.  Do we expect this to take more than O(10) iterations?  If not, why not set both `max_it` and `restart` to a smaller value like 30?

tests/test_stokes.py
```diff
@@ -106,15 +132,39 @@ def NSETest(butch, stage_type, splitting):
...
+        "snes_type": "newtonls",
+        "snes_converged_reason": None,
+        "snes_linesearch_type": "l2",
+        "snes_rtol": 1.e-8,
+        "ksp_rtol": 1.e-10,
+        "ksp_atol": 1.e-13,
+        "ksp_type": "fgmres",
+        "ksp_monitor": None,
+        "ksp_max_it": 200,
+        "ksp_gmres_restart": 30,
```
ScottMacLachlan, line 148: Same as above

ScottMacLachlan (approved):

PR #83, opened 2024-02-29
Fix ODE BCs

ScottMacLachlan:
As written in the Irksome paper, ODE BCs are supposed to set Dirichlet BCs on the stages to match the time derivative of the given Dirichlet BC on the solution.  This was unintentionally changed in #24, when we combined code between ODE and DAE BCs.

This PR fixes the constants introduced in #24 to be zero, so that we restore the original behaviour.


rckirby:
Thanks for the sleuthwork on this!

PR #84, opened 2024-02-29
Add test for ODE/DAE with incompatible BC

rckirby:
We should have added this with the previous PR fixing the BC.  But there you have it.


ScottMacLachlan (approved):
Looks good - thanks!

PR #87, opened 2024-05-23
Add adaptive timestepping

ScottMacLachlan:
This (re-)introduces code that allows for error estimation and adaptive timestepping, including a demo and tests.


demos/heat/demo_heat_adapt.py.rst
```diff
@@ -0,0 +1,126 @@
...
+  luparams = {"mat_type": "aij",
+              "ksp_type": "preonly",
+              "pc_type": "lu"}
+
+Most of Irksome's magic happens in the :class:`.TimeStepper`.  It
+transforms our semidiscrete form `F` into a fully discrete form for
+the stage unknowns and sets up a variational problem to solve for the
+stages at each time step.::
+
+  stepper = TimeStepper(F, butcher_tableau, t, dt, u, bcs=bc,
```
rckirby, line 107: This seems very clean.  Perhaps some comment in this demo as to what the parameters do?

demos/heat/demo_heat_adapt.py.rst
```diff
@@ -0,0 +1,126 @@
...
+			tol=1e-3, KI=1/15, KP=0.13)
+
+This logic is pretty self-explanatory.  We use the
+:class:`.TimeStepper`'s :meth:`~.TimeStepper.advance` method, which solves the variational
+problem to compute the Runge-Kutta stage values and then updates the solution.::
+
+  while (float(t) < float(Tf)):
+      if (float(t) + float(dt) > float(Tf)):
+          dt.assign(float(Tf) - float(t))
+      (adapt_error, adapt_dt) = stepper.advance()
```
rckirby, line 118: It looks like we don't do anything with `adapt_error`.  I suppose it's being used internally together with the other parameters to determine `adapt_dt`?

irksome/ButcherTableaux.py
```diff
@@ -17,14 +17,20 @@ class ButcherTableau(object):
     :arg c: a 1d array containing weights at which time-dependent
             terms are evaluated.
     :arg order: the (integer) formal order of accuracy of the method
+    :arg embedded_order: If present, the (integer) formal order of
+            accuracy of the embedded method
+    :arg gamma0: If present, the weight on the explicit term in the
+            embedded lower-order method
     """
-    def __init__(self, A, b, btilde, c, order):
+    def __init__(self, A, b, btilde, c, order, embedded_order, gamma0):
```
rckirby, line 25: It looks like this assumes that every `ButcherTableau` now has an embedded method.  Maybe we need an intermediate class like `EmbeddedButcherTableau`?

rckirby: I guess (seeing below) that you let `None` be an option for these.
Did you weigh subclassing versus checking for `None`?

irksome/stepper.py
```diff
@@ -76,6 +84,28 @@ def TimeStepper(F, butcher_tableau, t, dt, u0, **kwargs):
             F, butcher_tableau, t, dt, u0, bcs, appctx=appctx,
             solver_parameters=solver_parameters, nullspace=nullspace,
             bc_type=bc_type, splitting=splitting)
+    elif stage_type == "adapt":
+        bcs = kwargs.get("bcs")
+        bc_type = kwargs.get("bc_type", "DAE")
+        splitting = kwargs.get("splitting", AI)
```
rckirby, line 90: Do we need to consider allowing other types of BC, splittings, etc?  I get that this is a first stab.  It seems like maybe adaptivity is something that could be built on top of any of those approaches eventually.

irksome/stepper.py
```diff
@@ -265,3 +297,253 @@ def advance(self):
...
+            controls may adjust this before the step is taken.
+    :arg u0: A :class:`firedrake.Function` containing the current
+            state of the problem to be solved.
+    :arg tol: The temporal truncation error tolerance
+    :arg dtmin: Minimal acceptable time step.  An exception is raised
+            if the step size drops below this threshhold.
+    :arg dtmax: Maximal acceptable time step, imposed as a hard cap;
+            this can be adjusted externally once the time-stepper is
+            instantiated, by modifying `stepper.dt_max`
+    :arg KI: Integration gain for step-size controller.  Should be less
```
rckirby, line 326: `Integration gain` may not be a familiar term.  Quick description of it?

irksome/stepper.py
```diff
@@ -265,3 +297,253 @@ def advance(self):
...
+            then 1/p, where p is the expected order of the scheme
+    :arg KP: Proportional gain for step-size controller.
```
rckirby, line 328: Definition?

irksome/stepper.py
```diff
@@ -265,3 +297,253 @@ def advance(self):
...
+        if self.gamma0 != 0:
+            # Grab spaces for BCs
+            v = F.arguments()[0]
+            V = v.function_space()
+            num_fields = len(V)
+            num_stages = butcher_tableau.num_stages
+            btilde = butcher_tableau.btilde
+            ws = self.ws
+
+            class EmbeddedBCData(object):
```
rckirby, line 392: Is there any way we get code re-use from some other place where I've splatted out this wretched mess?

irksome/stepper.py
```diff
@@ -265,3 +297,253 @@ def advance(self):
...
+        # Initialize e to be gamma*h*f(old value of u)
+        error_func = Function(u0.function_space())
+        # Only do the hard stuff if gamma0 is not zero
+        if self.gamma0 != 0.0:
+            error_test = TestFunction(u0.function_space())
+            f_form = inner(error_func, error_test)*dx-self.gamma0*dtc*self.dtless_form
+            for gdat, gcur, gmethod in self.gblah:
+                gmethod(gcur, self.u0)
+            f_problem = NLVP(f_form, error_func, bcs=self.embbc)
+            solver_params = {
```
rckirby, line 475: Does this fail in parallel (are we defaulting the LU solver to PETSc, which is serial)?  And is there a way to override/select better options?  Looks like just a mass matrix, so probably CG with "something" could be made to work if we want to plumb in extra kwarg for these solver options.

tests/test_adaptive.py
```diff
@@ -0,0 +1,159 @@
...
+        stepper.advance()
+        t.assign(float(t) + float(dt))
+
+    u1, u2 = w.subfunctions
+    u1.interpolate(uexact_1)
+    u2.interpolate(uexact_2)
+    return norm(u1-uexact_1) + norm(u2-uexact_2)
+
+
+@pytest.mark.parametrize('N', [2**j for j in range(2, 4)])
```
rckirby, line 141: These tests make sure we get the right error.  Do we also need performance oriented tests to make sure we're actually getting something (fewer time steps, etc) out of doing adaptivity?

rckirby (changes requested):

demos/heat/demo_heat_adapt.py.rst
ScottMacLachlan: Good call.  I've added some text both here and around the adaptive loop.

demos/heat/demo_heat_adapt.py.rst
ScottMacLachlan: That's right.  This is now noted in the text above the loop.  I also tidied up the loop a bit, since the demo was written before some of the code was finalized.

irksome/ButcherTableaux.py
ScottMacLachlan: I didn't see a clean way to do this.  All of the schemes we currently have implemented are _essentially_ of type `CollocationButcherTableau` (because there's a general rule there, and we borrow from that case for LobattoIIIC).  So, we check that a `btilde` exists instead...  Happy to do this another way, but I don't currently see advantages or disadvantages to what is done now

irksome/stepper.py
ScottMacLachlan: On the one hand, I'm totally in agreement here.  On the other hand, this doesn't seem like the best place for a tutorial on [PID controllers](https://en.wikipedia.org/wiki/Proportional–integral–derivative_controller).  Suggestions for how to navigate this?  In essence, the step-size controller that we're using is from the literature [here](https://link.springer.com/article/10.1007/BF01934091).  These are the two knobs to twiddle, who default to recommended values from that paper, but I'm not sure how to convey insight into their choices in a doctoring...

irksome/stepper.py
ScottMacLachlan: I completely agree that we're recoding the same trick in several cases now.  This might be the worst place to look for better reuse, since it applies only to the case where we have an explicit first stage in an embedded scheme, which is not a case we have coded up elsewhere.

tests/test_adaptive.py
ScottMacLachlan: This is easy enough to do.  Are you thinking of a pure regression test, where we run to steady-state (with and without adaptive) and just confirm the iteration counts, or something more sophisticated?

irksome/stepper.py
ScottMacLachlan: This is just copied from the `deriv` case.  Just setting defaults here, not limiting selections, I think.

irksome/stepper.py
ScottMacLachlan: Good point - I had meant to do this, but forgotten.  Will plumb in options to select, but what do we have in mind as _good_ defaults here?

rckirby: Whatever firedrake's default linear solver is?  Or CG + ASM/ICC on blocks should be ok on a mass matrix?

tests/test_adaptive.py
rckirby: Nothing sophisticated, but maybe we'd like to test the number of time steps versus the number of uniform-sized time steps to get a comparable error?

rckirby (changes requested):
Several points are resolved.  I think the main issue to discuss is whether `stage_type` should include `adapt` or adaptivity should be included as a separate `kwarg`.  Also, still need to fix solver options (plumbing + default) for the error estimator.

irksome/stepper.py
ScottMacLachlan: I took another shot at this, with a bit more info and a reference.  Thoughts?

tests/test_adaptive.py
ScottMacLachlan: I've added a comparison test.  Solves a manufactured solution problem and compares that the adaptive time stepper is no more than 10% worse than fixed time steps (for reasonable choice of tolerance) and gets there in fewer steps.  Seem reasonable?

irksome/stepper.py
rckirby: Sure.  I think this is fine.

tests/test_adaptive.py
rckirby: I think this is a fine idea.

rckirby (approved):
Thanks!  This is a big feature addition.

PR #88, opened 2024-07-02
Rckirby/vandermonde

rckirby:
The stage-value formulation of RK methods solves for the coefficients of the collocation polynomial in the Lagrange basis associated with the collocation points.  This PR provides plumbing (and one specific example -- Bernstein) to solve for the coefficients in an alternative basis.  This gives rise to a more complicated linear system, but opens up opportunities to enforce bounds constraints uniformly in time by using Bernstein (or later, Chebyshev?) polynomials.


rckirby:
The math is not quite right here.  I need to fix it and reopen later.

PR #90, opened 2024-07-27
Rckirby/explicit

rckirby:
This PR does two things:
1) It adds an optimized explicit time stepper (which inherits from DIRK since it's an SDIRK).  One accesses this with `stage_type="explicit"` instead of `stage_type="dirk"`.  It automatically sets up SNES to never rebuild the jacobian or preconditioner, which is an optimization that works for RK.  So users just have to provide mass matrix inversion parameters.
2) It also adds a new family of "pseudo-energy-conserving"  explicit RK methods from de Leon, Ketcheson, and Ranocha.


demos/mixed_wave/demo_RTwave_PEP.py.rst
```diff
@@ -0,0 +1,112 @@
...
+   
+   p_t + \nabla \cdot u & = 0
+
+together with homogeneous Dirichlet boundary conditions
+
+.. math::
+
+   p = 0 \quad \textrm{on}\ \Gamma
+
+In this form, at each time, :math:`u` is a vector-valued function in the Soboleve space :math:`H(\mathrm{div})` and `p` is a scalar-valued function.  If we select appropriate test functions :math:`v` and :math:`w`, then we can arrive at the weak form
```
ScottMacLachlan, line 21: Fix spelling of Sobolev

demos/mixed_wave/demo_RTwave_PEP.py.rst
```diff
@@ -0,0 +1,112 @@
...
+  params = {"snes_type": "ksponly",
+            "ksp_type": "cg",
+            "pc_type": "icc"}
+
+  stepper = TimeStepper(F, butcher_tableau, t, dt, up0,
+                        stage_type="explicit",
+                        solver_parameters=params)
+
+
+And, as with the heat equation, our time-stepping logic is quite simple.  At easch time step, we print out the energy in the system::
```
ScottMacLachlan, line 92: Fix spelling of "each"

demos/mixed_wave/demo_RTwave_PEP.py.rst
```diff
@@ -0,0 +1,112 @@
...
+      if float(t) + float(dt) > 1.0:
+          dt.assign(1.0 - float(t))
+
+      stepper.advance()
+
+      t.assign(float(t) + float(dt))
+      print("{0:1.1e} {1:5e}".format(float(t), assemble(E)))
+
+If all is well with the world, the energy will be nearly identical (up
+to roundoff error) at each time step because the GL methods are
```
ScottMacLachlan, line 107: Probably we should cut/edit this text, since we're not using GL in this demo

tests/test_split.py
ScottMacLachlan, line 1: I think we don't want these changes

tests/test_wave_energy.py
```diff
@@ -11,7 +11,7 @@
 # various time steppers.
 
 
-def wave(n, deg, butcher_tableau, splitting=AI):
+def wave(n, deg, butcher_tableau, splitting):
```
ScottMacLachlan, line 14: This is a reasonable change, but not really related to this PR...

ScottMacLachlan (commented):
I would love to see a test or two focused on the explicit time steppers.  Is there a reason to not expand `test_wave_energy.py` to run these schemes through their paces?

demos/mixed_wave/demo_RTwave_PEP.py.rst
rckirby: done

demos/mixed_wave/demo_RTwave_PEP.py.rst
rckirby: done, good catch.

tests/test_split.py
rckirby: Right, they were an attempt to get CI to pass.  But we've skipped this test on the master branch.

tests/test_wave_energy.py
rckirby: The options to `wave` have been reworked a good bit to allow us to test GL and PEPRK methods with the same suite.

rckirby:
> I would love to see a test or two focused on the explicit time steppers. Is there a reason to not expand `test_wave_energy.py` to run these schemes through their paces?

Yes, good point.  I've done this now.

demos/mixed_wave/demo_RTwave_PEP.py.rst
```diff
@@ -0,0 +1,112 @@
+:orphan:
```
ScottMacLachlan, line 1: Why do we need this tag?

Unrelated, we should probably edit `docs/source/index.rst` to add this demo to the index

demos/mixed_wave/demo_RTwave_PEP.py.rst
```diff
@@ -0,0 +1,112 @@
...
+stages than classical explicit methods but have much better energy conservation.::
+
+  butcher_tableau = PEPRK(4, 2, 5)
+
+Like the heat equation demo, we are just using a direct method to solve the
+system at each time step::
+
+  params = {"snes_type": "ksponly",
+            "ksp_type": "cg",
+            "pc_type": "icc"}
```
ScottMacLachlan, line 85: The text and `params` don't agree.  This is using CG preconditioned by ICC, not a direct method.

tests/test_split.py
ScottMacLachlan, line 1: I'm still seeing changes in this file...

ScottMacLachlan (approved):

tests/test_split.py
rckirby: The changes are in a different place.  Putting it back, I just put both back to mumps.

PR #91, opened 2024-08-13
skip the pernicious split NSE test

rckirby:
What the title says.  This one test is of a feature we're not actively developing, unsure of, plan to do future work, and keep having to deal with asynchronous solver failures of.


ScottMacLachlan (approved):
LGTM

PR #92, opened 2024-08-14
refactor WSO dirk interface to all be in one class; update dirk tests…

rckirby:
We're going to do for WSO dirks what we recently did for PEP RK methods: put them all inside a single class that takes the stage & other parameters as inputs.


demos/dirk/demo_dirk_parameters.py.rst
```diff
@@ -1,31 +1,13 @@
-Using solver options to make gain efficiency for DIRK methods
+Accessing DIRK methods
 =============================================================
 Many practitioners favor diagonally implicit methods over fully implicit
 ones since the stages can be computed sequentially rather than concurrently.
+We support a range of DIRK methods and provide
```
ScottMacLachlan, line 5: Sentence fragment...

irksome/wso_dirk_tableaux.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+class WSODIRK(ButcherTableau):
+    def __init__(self, ns, order, ws_order):
+        try:
+            A, b, c = wsodict[ns, order, ws_order]
+        except KeyError:
+            raise NotImplementedError("No WSO DIRK method for that combination of stages, order and weak stage order")
+        self.ws_order = ws_order
+        super(WSODIRK, self).__init__(A, b, None, c, order)
```
ScottMacLachlan, line 183: This will need updating if we merge the adaptive PR in its current form, to pass `None` for `Atilde` and `btilde`

ScottMacLachlan (approved):
Looks quite reasonable to me!

irksome/wso_dirk_tableaux.py
rckirby: Good catch, we'll update it on the adaptive PR end after I get that reviewed.

PR #96, opened 2024-08-20
Rckirby/vandermonde

rckirby:
Make Vandermonde formulation work for collocation methods.
Also, plumb in options to do bounds-constrained time stepping.


irksome/stage.py
```diff
@@ -361,14 +418,70 @@ def __init__(self, F, butcher_tableau, t, dt, u0, bcs=None,
...
+            bounds_type, lower, upper = bounds
+            lower_func = Function(V)
+            upper_func = Function(V)
+            if not isiterable(lower):
+                lower = (lower,)
+            if not isiterable(upper):
+                upper = (upper,)
+            for l, lf in zip(lower, lower_func.subfunctions):
+                if l is not None:
+                    lf.assign(l)
```
ScottMacLachlan, line 440: Are these `assign`s unavoidable?  Not that we've made much progress on being adjoint-able, but this seems bad from that perspective...

tests/test_split.py
```diff
@@ -196,8 +196,9 @@ def Ffull(z, test):
 @pytest.mark.skip(reason="solver parameters :(")
 @pytest.mark.parametrize('N', [16])
 @pytest.mark.parametrize('num_stages', [2])
-@pytest.mark.parametrize('Fimp,Fexp', [(FimpSt, FexpSt),
-                                       (FimpLI, FexpLI)])
+@pytest.mark.parametrize('Fimp,Fexp', [(FimpLI, FexpLI)])
+# @pytest.mark.parametrize('Fimp,Fexp', [(FimpSt, FexpSt),
+#                                        (FimpLI, FexpLI)])
```
ScottMacLachlan, line 201: How is this related?

irksome/tools.py
```diff
@@ -117,3 +118,43 @@ def __init__(self, msh):
 
     def Constant(self, val=0.0):
         return Function(self.V).assign(val)
+
+
+def ConstantOrZero(x, MC):
+    return Zero() if abs(complex(x)) < 1.e-10 else MC.Constant(x)
+
+
+def bc2space(bc, V):
```
ScottMacLachlan, line 127: I agree this is a useful function to add, but we don't seem to use it anywhere yet...

tests/test_accuracy.py
```diff
@@ -56,11 +52,23 @@ def heat(n, deg, time_stages, stage_type="deriv", splitting=IA):
...
+    conv = np.log2(diff[:-1] / diff[1:])
+    assert (conv > convrate).all()
+
+
+@pytest.mark.parametrize(('deg', 'convrate', 'time_stages'),
+                         [(2, 2.8, i) for i in (2, 3)]
+                         + [(3, 3.8, i) for i in (3, 4)])
+def test_heat_bern(deg, convrate, time_stages):
+    diff = np.array([heat(i, deg, time_stages, **{"stage_type": "value", "basis_type": "Bernstein"}) for i in range(3, 8)])
+    print(diff)
```
ScottMacLachlan, line 71: These `print` statements should be removed for tests

tests/test_bern.py
```diff
@@ -0,0 +1,167 @@
...
+    rhs = expand_derivatives(diff(pexact, t) + div(uexact))
+
+    F = (inner(Dt(p), w) * dx
+         + inner(div(u), w) * dx
+         - inner(rhs, w) * dx
+         + inner(u, v) * dx
+         - inner(p, div(v)) * dx
+         + inner(pexact, dot(v, n)) * ds)
+
+    up.subfunctions[1].project(pexact)
```
ScottMacLachlan, line 100: Why is `up.subfunctions[0]` not initialized?

tests/test_bern.py
```diff
@@ -0,0 +1,167 @@
...
+    print(conv)
+    assert (conv > (deg-0.1)).all()
+
+
+@pytest.mark.parametrize('num_stages', (2, 3))
+@pytest.mark.parametrize('bounds_type', ('stage', 'last_stage'))
+@pytest.mark.parametrize('basis_type', ('Bernstein', None))
+def test_mixed_heat_bern_bounds(num_stages, bounds_type, basis_type):
+    deg = 1
+    bounds = (bounds_type, (None, 0), (None, None))
```
ScottMacLachlan, line 136: This remains quite cryptic to me.  I think the right thing to do is have a fully commented demo that shows how to set bounds and use them...

ScottMacLachlan (changes requested):

tests/test_bern.py
danshapero: I think strictly speaking it doesn't matter, this sets the initial pressure field and the solver will compute a velocity field in the next timestep. The comparison in lines 113-114 is just against the final value of the exact and numerical solutions; if the test instead used, say, the time integral of the discrepancy then you'd be testing things incorrectly by not setting `up.subfunctions[0]` as well.

danshapero (dismissed):
Maybe worth discussing more what the API should be for setting bounds constraints.

irksome/stage.py
rckirby: I thought that `dat` bad, `assign` good?

irksome/stage.py
```diff
@@ -109,20 +124,47 @@ def getFormStage(F, butch, u0, t, dt, bcs=None, splitting=None,
...
     vecconst = np.vectorize(lambda c: MC.Constant(c))
+
     C = vecconst(butch.c)
     A = vecconst(butch.A)
 
+    veccorz = np.vectorize(lambda c: ConstantOrZero(c, MC))
+    Vander = veccorz(vandermonde)
+    Vander_inv = veccorz(np.linalg.inv(vandermonde))
+
+    # multiply [u0bits; UUbits] = Vander @ [u0bits; UUbits]
```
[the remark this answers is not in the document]
rckirby: Because of the structure of the Bernstein Vandermonde matrix, the first entry is unchanged, and plus it's not part of the linear system we're solving for.

tests/test_split.py
rckirby: Yet Another Random Stokes Failure.

irksome/tools.py
rckirby: Right, I had thought I needed it at one point, hadn't removed.

irksome/stage.py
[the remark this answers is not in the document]
rckirby: I'm expanding the comments a bit more.

tests/test_bern.py
rckirby: Right, there isn't an official initial condition for the flux variable for the PDE, but it doesn't hurt anything to set the initial flux also (I just checked).

tests/test_bern.py
rckirby: Good point, I'll double check order against Firedrake/PETSc.  We can probably unpack the `None` option.  We will have a documented demo in before we try to merge.

tests/test_bern.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+def mixed_heat(n, deg, butcher_tableau, solver_parameters,
+               **kwargs):
+    N = 2**n
+    msh = UnitSquareMesh(N, N)
+
+    V = FunctionSpace(msh, "RT", deg)
+    el_type = "Bernstein" if deg > 1 else "DG"
+    W = FunctionSpace(msh, el_type, deg-1)
```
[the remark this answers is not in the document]
rckirby: Oh, yes.  Correct.

tests/test_accuracy.py
rckirby: Done

tests/test_bern.py
rckirby: Yes, the order is fixed and we are using lower, upper now.

rckirby: And there are fewer options now  -- we just have `upper` and `lower` (like for the NLVS).  If you want to constrain one piece of a mixed space, it's on you to set the bounds on the other sub functions to be INFINITY or NINFINITY.

demos/bounded_heat/bounded_heat.py.rst
```diff
@@ -0,0 +1,202 @@
...
+    Tf = MC.Constant(1.0)
+
+We will find an approximate solution at time :math:`t=1.0` with and without 
+enforcing a constraint on the lower bound. We will need the following pair of solver 
+parameters: ::
+
+    lu_params = {
+        "snes_type": "ksponly",
+        "ksp_type": "preonly",
+        "mat_typ": "aij",
```
rckirby, line 65: There is a typo here: `mat_type`.  Also, for parallel structure with `vi_params`, we should set it in both places or neither.

demos/bounded_heat/bounded_heat.py.rst
```diff
@@ -0,0 +1,202 @@
...
+
+    with upper.dat.vec as upper_vec:
+        upper_vec.set(np.inf)
+
+    with lower.dat.vec as lower_vec:
+        lower_vec.set(0.0)
+
+    nlvs.solve(bounds=(lower, upper))
+
+    u_c = u.copy(deepcopy=True)
```
rckirby, line 111: I think `u_c.assign(u)` is cleaner?

demos/bounded_heat/bounded_heat.py.rst
```diff
@@ -0,0 +1,202 @@
...
+
+For the unconstrained approximation, we configure the :class:`.TimeStepper` in a 
+familiar way: ::
+
+    stepper = TimeStepper(F, butcher_tableau, t, dt, u, bcs=bc, solver_parameters=lu_params)
+
+We will enforce nonnegativity when finding the constrained approximation. We now set up the keyword database to 
+configure an instance of :class:`.TimeStepper` for this task. We first specify, using the 
+keyword ``stage_type``, that we wish to use a stage-value formulation of the underlying collocation 
+method. The keyword ``basis_type`` then allows us to change the basis of the collocation
```
rckirby, line 139: A comment that `basis_type` not being Bernstein will ensure positivity at the discrete stages (and time levels) but not uniformly between them seems in order.  We can apply bounds in either form, we just get a different method.

demos/bounded_heat/bounded_heat.py.rst
```diff
@@ -0,0 +1,202 @@
...
+We set the bounds as follows: ::
+
+    lb = Function(V)
+    lb.assign(0)
+
+    ub = None
+
+    bounds = ('stage', lb, ub)
+
+Passing the :meth:`~.TimeStepper.advance` method ``bounds`` will enforce the
```
rckirby, line 160: This is not (quite) right.  There is no `TimeStepper` class since that is a convenience function that takes the keyword arguments and constructs a class of some type.  In fact, `advance` only supports the `bounds` argument if we are in stage value form (Bernstein or standard is fine).

demos/bounded_heat/bounded_heat.py.rst
```diff
@@ -0,0 +1,202 @@
...
+        if minv_c < 0:
+            mins_c.append((float(t), minv_c))
+
+        print(float(t))
+  
+Finally, we print the relative :math:`L^2` error and the time and severity (if any) of constraint violations: ::
+
+    np.set_printoptions(legacy='1.25')
+
+    print()
```
rckirby, line 197: I recommend printing some strings as well saying what we're looking at, like
```print(f"Relative L^2 norm of unconstrained solution: {errornorm(uexact, u) / norm(uexact)}"```.

File: demos/bounded_heat/bounded_heat.py.rst [hunk 14]
```diff
@@ -0,0 +1,202 @@
...
+    print(norm(u - uexact) / norm(uexact))
+    print(norm(u_c - uexact) / norm(uexact))
+    print()
+    print(mins)
```
rckirby, line 201: Also, the output is probably more descriptive/readable if you print out the discrete times as well, and only print the time levels where the bounds are violated.

demos/bounded_heat/bounded_heat.py.rst
```diff
@@ -0,0 +1,202 @@
...
+    print(mins_c)
```
rckirby, line 202: And also print out the same thing (with a hopefully empty list!) for the constrained problems.

demos/bounded_heat/bounded_heat.py.rst
```diff
@@ -0,0 +1,202 @@
...
+    msh = UnitSquareMesh(N, N)
+    V = FunctionSpace(msh, "Bernstein", 2)
+
+In order to enforce bounds constraints in time, we must utilize a collocation method. 
+In this demo, we will time-step using the L-stable, fully implicit, 2-stage RadauIIA 
+Runge-Kutta method. The bounds will be passed as an argument to the 
+:meth:`~.TimeStepper.advance` method. We now define the Butcher Tableau and variables to store the 
+time step, current time, and final time: ::
+
+    butcher_tableau = RadauIIA(2)
```
rckirby, line 51: Important note: I think that RadauIIA(1)/backward Euler will guarantee the bounds constraints, but higher-order methods don't.  This is worth putting in the discussion.

demos/bounded_heat/bounded_heat.py.rst
```diff
@@ -0,0 +1,202 @@
...
+
+We also need some UFL tools in order to manufacture a solution: ::
+
+    from ufl.algorithms import expand_derivatives
+
+Finally, numpy provides us with the upper bound of infinity: ::
+
+    import numpy as np
+
+We first define the mesh and the necessary function space. We choose
```
rckirby, line 37: Expand the comment on Bernstein: constraining the degrees of freedom for the Bernstein basis guarantees positivity of the solution, but this is not true for Lagrange.  The nodal values there can be constrained, but this doesn't guarantee uniform positivity.

demos/bounded_heat/bounded_heat.py.rst
jdsteph2: While the exact solution chosen is positive, the manufactured RHS is not.  Unfortunately, RadauIIA(1) does not provide a positive approximation in this case.  The same initial condition with no forcing would also work (and backward Euler would be positive), but then there would be no exact solution to compare the approximations against.  I can find a more suitable example if we want to highlight the difference between backward Euler/higher-order methods.

rckirby: This might not be wrong -- the order barrier says that higher-order can't preserve bounds, not that backward Euler will for any problem.  I think we're fine here.

irksome/stage.py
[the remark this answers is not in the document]
rckirby: And the commentary was wrong, anyway.

tests/test_bern.py
rckirby: This is pretty fully reworked.  See updated tests and also john's committed demo.

irksome/stage.py
```diff
@@ -252,85 +275,113 @@ def getFormStage(F, butch, u0, t, dt, bcs=None, splitting=None,
...
+                else:
+                    Vsp = V.sub(sub)
+
+            bcarg = as_ufl(bc._original_arg)
+            try:
+                gdat = assemble(interpolate(bcarg, Vsp))
+                gmethod = lambda gd, gc: gd.interpolate(gc)
+            except:  # noqa: E722
+                gdat = project(bcarg, Vsp)
+                gmethod = lambda gd, gc: gd.project(gc)
```
ScottMacLachlan, line 341: As in the demo, I think we need some solver options for this projection in the case where we want to satisfy bounds constraints.

tests/test_stokes.py
```diff
@@ -196,7 +196,7 @@ def NSETest(butch, stage_type, splitting):
 @pytest.mark.parametrize('butch', (LobattoIIIC, RadauIIA))
 def test_Stokes(N, butch, time_stages, stage_type, splitting):
     error = StokesTest(N, butch(time_stages), stage_type, splitting)
-    assert abs(error) < 1e-8
+    assert abs(error) < 3e-8
```
ScottMacLachlan, line 199: I'm generally opposed to tweaking unrelated tests as part of a PR, but I see the efficiency in this...

tests/test_dirk.py
```diff
@@ -290,4 +288,4 @@ def test_stokes_bcs(butcher_tableau, bctype):
         stepper.advance()
         stepperdirk.advance()
         t.assign(float(t) + float(dt))
-        assert (errornorm(u_dirk, u) / norm(u)) < 1.e-10
+        assert errornorm(u_dirk, u) < 1.e-7
```
ScottMacLachlan, line 291: I'm generally opposed to tweaking unrelated tests as part of a PR, but I see the efficiency in this...

demos/bounded_heat/demo_bounded_heat.py.rst
```diff
@@ -0,0 +1,227 @@
...
+
+    uexact = 0.5 * exp(-t) * (1 + (tanh((0.1 - sqrt((x - 0.5) ** 2 + (y - 0.5) ** 2)) / 0.015)))
+
+    rhs = expand_derivatives(diff(uexact, t)) - div(grad(uexact))
+
+Note that the exact solution is uniformly positive in space and time. Using a manufactured 
+solution, one would typically project the exact solution at time :math:`t = 0` onto the 
+approximation space and use the projection as the initial condition. In this case, the 
+projection does not satisfy the lower bound of :math:`0`. We instead solve a variational inequality 
+to find a bounds-preserving initial condition: ::
```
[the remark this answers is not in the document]
rckirby: Good catch.  Our BC projection for this particular demo seems to get the BC's right, but I've put in some plumbing to make sure of that and updated the demo.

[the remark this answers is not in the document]
rckirby: Interpolation doesn't work for Bernstein, and even if it did, it couldn't preserve the bounds due to the order barrier of Nochetto/Wahlbin.

demos/bounded_heat/demo_bounded_heat.py.rst
```diff
@@ -0,0 +1,227 @@
...
+
+    F = (inner(Dt(u), v) * dx + inner(grad(u), grad(v)) * dx - inner(rhs, v) * dx)
+
+    v_c = TestFunction(V)
+
+    F_c = (inner(Dt(u_c), v_c) * dx + inner(grad(u_c), grad(v_c)) * dx - inner(rhs, v_c) * dx)
+
+We use exact boundary conditions in both cases: ::
+
+    bc = DirichletBC(V, uexact, "on_boundary")
```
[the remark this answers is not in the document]
rckirby: Good catch, and we've updated the BC mechanism to enable one to pass in constraints.  In fact, this particular example didn't trigger that potential failure mode.

demos/bounded_heat/demo_bounded_heat.py.rst
```diff
@@ -0,0 +1,227 @@
...
+
+Note that if one does not set the ``basis_type`` to Bernstein, the standard basis will be used.  The bounds will 
+then be enforced at the discrete stages and time levels, but not uniformly between them. 
+
+We set the bounds as follows: ::
+
+    lb = Function(V)
+    lb.assign(0)
+
+    ub = None
```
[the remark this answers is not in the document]
rckirby: I changed it so that we just reused the bounds from the IC for the time stepping.

irksome/stage.py
rckirby: Extra logic is added.

tests/test_dirk.py
rckirby: They were failing locally on my Mac (of course).  Point taken.

demos/bounded_heat/demo_bounded_heat.py.rst
```diff
@@ -0,0 +1,227 @@
...
+the Bernstein basis).
+
+We now advance both semidiscrete systems in the usual way. We add the bounds as an argument 
+to the :meth:`~.TimeStepper.advance` method for the constrained approximation.  
+
+In order to monitor our approximate solutions, we check the minimum value of each after every step in time. 
+If an approximate solution violates the lower bound, we append a tuple to indicate the time and minimum value. ::
+
+    vios = []
+    vios_c = []
```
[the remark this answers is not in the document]
rckirby: done.

ScottMacLachlan (approved):
Looks like this is almost there!

demos/bounded_heat/demo_bounded_heat.py.rst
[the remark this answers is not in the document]
jdsteph2: Agreed--foreshadowing added.

demos/bounded_heat/demo_bounded_heat.py.rst
```diff
@@ -0,0 +1,226 @@
...
+
+for some known functions :math:`f` and :math:`g`. The solution will be some function :math:`u\in V`, for 
+a suitable function space :math:`V`.
+
+The weak form is found by multiplying by an arbitrary test function :math:`v\in V` and integrating over :math:`\Omega`. 
+We then have the variational problem of finding :math:`u:[0,T]\rightarrow V` such that 
+.. math::
+
+    (u_t, v) + (\nabla u, \nabla v) = (f, v)
+
```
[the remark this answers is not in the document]
jdsteph2: Ah, very fair.  Thank you.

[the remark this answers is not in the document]
ScottMacLachlan: Looks good!

demos/bounded_heat/demo_bounded_heat.py.rst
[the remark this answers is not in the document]
ScottMacLachlan: LGTM

Alaabasheer09:
i want code to building in matlab by Runge-Kutta

PR #97, opened 2024-10-03
Specify that build runners run Linux

dham:
This is in preparation for the rollout of Mac build runners.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ on:
 jobs:
 
   test:
-    runs-on: self-hosted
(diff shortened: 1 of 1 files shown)


rckirby:
Will we need to change this later to make sure we also run on Mac?

dham:
> Will we need to change this later to make sure we also run on Mac?

No. This is just about testing. We don't need to test Irksome on different hardware because it's sufficiently high in the mathematical stack that it's exceptionally unlikely to matter what hardware it's on. The Mac builders are being introduced so we can detect Mac compilation failures for the lower-level components of Firedrake.

PR #98, opened 2024-10-14
Fix boundary conditions

Homingdung:
To allow `DirichletBC(V, 0, ...) `, the special case requires no computation for stage boundary condition, everything should be zero

### DIFF
--- a/irksome/getForm.py
+++ b/irksome/getForm.py
@@ -16,43 +16,32 @@ from .deriv import TimeDerivative  # , apply_time_derivatives
 class BCStageData(object):
     def __init__(self, V, gcur, u0, u0_mult, i, t, dt):
         if gcur == 0:  # special case DirichletBC(V, 0, ...), do nothing
-            gdat = gcur
-            gmethod = lambda g, u: None
+            u0 = 0
         elif V.component is not None:     # bottommost space is bit of VFS
             if V.parent.index is None:  # but not part of a MFS
                 sub = V.component
-                try:
-                    gdat = assemble(interpolate(gcur-u0_mult[i]*u0.sub(sub), V))
-                    gmethod = lambda g, u: gdat.interpolate(g-u0_mult[i]*u.sub(sub))
-                except NotImplementedError:
-                    gdat = project(gcur-u0_mult[i]*u0.sub(sub), V)
-                    gmethod = lambda g, u: gdat.project(g-u0_mult[i]*u.sub(sub))
+                u0 = u0.sub(sub)
             else:   # V is a bit of a VFS inside an MFS
                 sub0 = V.parent.index
                 sub1 = V.component
-                try:
-                    gdat = assemble(interpolate(gcur-u0_mult[i]*u0.sub(sub0).sub(sub1), V))
-                    gmethod = lambda g, u: gdat.interpolate(g-u0_mult[i]*u.sub(sub0).sub(sub1))
-                except NotImplementedError:
-                    gdat = project(gcur-u0_mult[i]*u0.sub(sub0).sub(sub1), V)
-                    gmethod = lambda g, u: gdat.project(g-u0_mult[i]*u.sub(sub0).sub(sub1))
+                u0 = u0.sub(sub0).sub(sub1)
         else:  # V is not a bit of a VFS
             if V.index is None:  # not part of MFS, either
-                try:
-                    gdat = assemble(interpolate(gcur-u0_mult[i]*u0, V))
-                    gmethod = lambda g, u: gdat.interpolate(g-u0_mult[i]*u)
-                except NotImplementedError:
-                    gdat = project(gcur-u0_mult[i]*u0, V)
-                    gmethod = lambda g, u: gdat.project(g-u0_mult[i]*u)
+                pass
             else:  # part of MFS
                 sub = V.index
-                try:
-                    gdat = assemble(interpolate(gcur-u0_mult[i]*u0.sub(sub), V))
-                    gmethod = lambda g, u: gdat.interpolate(g-u0_mult[i]*u.sub(sub))
-                except NotImplementedError:
-                    gdat = project(gcur-u0_mult[i]*u0.sub(sub), V)
-                    gmethod = lambda g, u: gdat.project(g-u0_mult[i]*u.sub(sub))
-
+                u0 = u0.sub(sub)
+        if u0 == 0:
+            gdat = gcur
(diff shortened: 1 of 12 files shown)


rckirby (changes requested):
Thanks for tackling this, it's long been a major annoyance.  However, I tried out your change elsewhere (in `test_stokes.py`) and find that that is somehow broken.  I expected stage value formulation to fail in that test since we haven't done similar updates to how that handles BC, but the classic stage-derivative form you attempted to fix is failing here.  Please take a look.

irksome/stage.py
```diff
@@ -50,6 +50,20 @@ def getBits(num_stages, num_fields, u0, UU, v, VV):
     return u0bits, vbits, VVbits, UUbits
 
 
+def gstuff(V, g):
+    if g == 0:
+        gdat = 0
```
pbrubeck, line 55:
```suggestion
        gdat = g
```

irksome/getForm.py
```diff
@@ -15,41 +15,33 @@
 
 class BCStageData(object):
     def __init__(self, V, gcur, u0, u0_mult, i, t, dt):
-        if V.component is not None:     # bottommost space is bit of VFS
+        if gcur == 0:  # special case DirichletBC(V, 0, ...), do nothing
+            u0 = 0
```
pbrubeck, line 19:
```suggestion
        if gcur == 0:  # special case DirichletBC(V, 0, ...), do nothing
            pass
```

irksome/getForm.py
```diff
@@ -15,41 +15,33 @@
...
-                try:
-                    gdat = assemble(interpolate(gcur-u0_mult[i]*u0.sub(sub), V))
-                    gmethod = lambda g, u: gdat.interpolate(g-u0_mult[i]*u.sub(sub))
-                except:  # noqa: E722
-                    gdat = project(gcur-u0_mult[i]*u0.sub(sub), V)
-                    gmethod = lambda g, u: gdat.project(g-u0_mult[i]*u.sub(sub))
-
+                u0 = u0.sub(sub)
+        if u0 == 0:
+            gdat = gcur
```
pbrubeck, line 35:
```suggestion
        if gcur == 0:
            gdat = gcur
```

irksome/getForm.py
```diff
@@ -15,41 +15,33 @@
...
+            gmethod = lambda g, u: None
+        else:
+            gexpr = gcur-u0_mult[i]*u0
+            try:
+                gdat = assemble(interpolate(gexpr, V))
+                gmethod = lambda g, u: gdat.interpolate(gexpr)
+            except (NotImplementedError, AttributeError):
+                gdat = project(gexpr, V)
+                gmethod = lambda g, u: gdat.project(gexpr)
```
pbrubeck, line 44: Are both `g` and `u` changing each time?

rckirby: In principle, they could be.  Looking at this now after a few years, it seems that `g` is unused in the lambda expression...

irksome/stage.py
```diff
@@ -271,44 +260,26 @@ def getFormStage(F, butch, u0, t, dt, bcs=None, splitting=None, vandermonde=None
...
             bcparams, bclower, bcupper = bc_constraints[bc]
-            for i in range(num_stages):
-                Vsp, Vbigi = stage2spaces4bc(bc, V, Vbig, i)
-                gdat = Function(Vsp)
-                vbc = TestFunction(Vsp)
-                gmethod = lambda gd, gc: solve(inner(gd - gc, vbc) * dx == 0,
-                                               gd, solver_parameters=bcparams)
             gcur = replace(bcarg, {t: t+C[i] * dt})
-            gblah_cur.append((gdat, gcur - Vander_col[i] * bcarg, gmethod))
+            gcur = gcur - Vander_col[i] * bcarg
```
rckirby, line 266: Is there a reason to prefer this over `gcur -= ...`?

irksome/stage.py
```diff
@@ -271,44 +260,26 @@ def getFormStage(F, butch, u0, t, dt, bcs=None, splitting=None, vandermonde=None
...
-
-                gcur = replace(bcarg, {t: t+C[i]*dt})
-                gblah_cur.append((gdat, gcur - Vander_col[i] * bcarg, gmethod))
-
             gdats_cur = np.zeros((num_stages,), dtype="O")
             for i in range(num_stages):
-                gdats_cur[i] = gblah_cur[i][0]
+                Vbigi = stage2spaces4bc(bc, V, Vbig, i)
+                gcur = replace(bcarg, {t: t+C[i]*dt})
+                gcur = gcur - Vander_col[i] * bcarg
```
rckirby, line 272: same here.

irksome/stage.py
```diff
@@ -271,44 +260,26 @@ def getFormStage(F, butch, u0, t, dt, bcs=None, splitting=None, vandermonde=None
     for bc in bcs:
         bcarg = as_ufl(bc._original_arg)
 
-        gblah_cur = []
-
         if bc in bc_constraints:
             bcparams, bclower, bcupper = bc_constraints[bc]
```
rckirby, line 264: `bcparams`, `bclower`, etc seem unused now?  The idea of this bit is that we need to be able to pass in special parameters to project the boundary condition expression into the finite element space subject to bounds constraints (e.g. using a VI solver).  It looks like this capability just disappeared.

rckirby (changes requested):
I think this is very very clean.  Except we have a technical corner case about no longer being able to do bounds-constrained projections for discrete boundary conditions.

irksome/stage.py
pbrubeck: Constants cannot be updated

ScottMacLachlan (commented):
Modulo answers to @rckirby 's questions about the bounds-constrained case, this all makes sense to me.  Seems much cleaner than what we were doing, and resolves a few things as well!  Many thanks!

rckirby:
Right now, our only use cases are constant upper and/or lower bounds.  I
think we should not make the perfect the enemy of the good.  For now, if
you have a `Function` in the upper or lower bound and update it before the
solver is called, the new bounds will be in place.  We don't have a
mechanism for `bounds` being an expression, but that's not an issue because
1) FD's bounds mechanism is the same and 2) you'd need a bounds-constrained
projection to pull it off..

irksome/stage.py
rckirby: Ah, I didn't think about that case.

rckirby (approved):

PR #99, opened 2024-10-17
Add new PC based on providing an alternate time-dependent PDE

rckirby:
It does what it says.  Tested by preconditioning reaction/diffusion with just diffusion as a simple example.


ScottMacLachlan (approved):
This all looks good to me!  Thanks!

PR #101, opened 2024-11-15
fix solver options for explicit solver

rckirby:
If one passed `solver_parameters=None` into an explicit stepper, it's rewriting the Jacobian/PC options should fail.  This converts the `None` into an empty dictionary.


ScottMacLachlan (approved):
LGTM.  Thanks!

PR #103, opened 2024-11-17
Fix DIRK boundary conditions

### LINKED ISSUE #102: Dirichlet BCs don't work with explicit
Strongly imposing Dirichlet BCs with `stage_type="explicit"` doesn't work.  Looking through the code, this seems obvious, since we have
```
                gdat /= dtc * float(AA[i, i])
```
[here](https://github.com/firedrakeproject/Irksome/blob/dc9f91cc9a1e4eac2f141e30a8de4400d4ef545f/irksome/dirk_stepper.py#L222C30-L222C46), which does bad things when `AA[i,i] == 0`.

Mathematically, I think we're just "off-by-one", in that our BC on `k[0]` should be derived from matching the BC at stage time `c[1]`, and so forth.  We need to think a bit about how to implement this in a non-intrusive way, though.

ScottMacLachlan:
When BCs were refactored in #98, the DIRK case was left untouched.

This PR aims to correct that, as well as addressing the root problem in #102, where we don't implement the correct DAE-style BCs for explicit schemes

fixes #102

### DIFF
--- a/irksome/dirk_stepper.py
+++ b/irksome/dirk_stepper.py
@@ -71,7 +71,7 @@ def getFormDIRK(F, ks, butch, t, dt, u0, bcs=None):
         new_bc = bc.reconstruct(g=gdat)
         bcnew.append(new_bc)
 
-    return stage_F, (k, g, a, c, a_vals, d_val), bcnew
+    return stage_F, (k, g, a, c), bcnew, (a_vals, d_val)
 
 
 class DIRKTimeStepper:
@@ -105,7 +105,7 @@ class DIRKTimeStepper:
         # that we update as we go.  We need to remember the
         # stage values we've computed earlier in the time step...
 
-        stage_F, (k, g, a, c, a_vals, d_val), bcnew = getFormDIRK(
+        stage_F, (k, g, a, c), bcnew, (a_vals, d_val) = getFormDIRK(
             F, self.ks, butcher_tableau, t, dt, u0, bcs=bcs)
 
         self.bcnew = bcnew
@@ -128,11 +128,13 @@ class DIRKTimeStepper:
             self.problem, appctx=appctx, solver_parameters=solver_parameters,
             nullspace=nullspace)
 
-        self.kgac = k, g, a, c, a_vals, d_val
+        self.kgac = k, g, a, c
+        self.bc_constants = a_vals, d_val
 
-    def update_bc_constants(self, i, a_vals, d_val, c):
+    def update_bc_constants(self, i, c):
         AAb = self.AAb
         CCone = self.CCone
+        a_vals, d_val = self.bc_constants
         ns = AAb.shape[1]
         for j in range(i):
             a_vals[j].assign(AAb[i, j])
@@ -142,7 +144,7 @@ class DIRKTimeStepper:
         c.assign(CCone[i])
 
     def advance(self):
-        k, g, a, c, a_vals, d_val = self.kgac
+        k, g, a, c = self.kgac
         ks = self.ks
         u0 = self.u0
         dtc = float(self.dt)
@@ -160,7 +162,7 @@ class DIRKTimeStepper:
                     gbit += dtc * float(AA[i, j]) * kbit
 
             # update BC constants for the variational problem
-            self.update_bc_constants(i, a_vals, d_val, c)
+            self.update_bc_constants(i, c)
             a.assign(AA[i, i])
 
             # solve new variational problem, stash the computed
--- a/irksome/dirk_stepper.py
+++ b/irksome/dirk_stepper.py
@@ -1,5 +1,5 @@
 import numpy
-from firedrake import DirichletBC, Function
+from firedrake import Function
 from firedrake import NonlinearVariationalProblem as NLVP
 from firedrake import NonlinearVariationalSolver as NLVS
 from firedrake import split
@@ -7,57 +7,7 @@ from ufl.constantvalue import as_ufl
 
 from .deriv import TimeDerivative
 from .tools import replace, MeshConstant
-
-
-class BCThingy:
-    def __init__(self):
-        pass
-
-    def __call__(self, u):
-        return u
-
-
-class BCCompOfNotMixedThingy:
-    def __init__(self, comp):
-        self.comp = comp
-
-    def __call__(self, u):
-        return u[self.comp]
-
-
-class BCMixedBitThingy:
-    def __init__(self, sub):
-        self.sub = sub
-
-    def __call__(self, u):
-        return u.sub(self.sub)
-
-
-class BCCompOfMixedBitThingy:
-    def __init__(self, sub, comp):
-        self.sub = sub
-        self.comp = comp
-
-    def __call__(self, u):
-        return u.sub(self.sub)[self.comp]
-
-
-def getThingy(V, bc):
-    num_fields = len(V)
-    Vbc = bc.function_space()
-    if num_fields == 1:
-        comp = Vbc.component
-        if comp is None:
-            return BCThingy()
-        else:
-            return BCCompOfNotMixedThingy(comp)
-    else:
-        sub = bc.function_space_index()
-        comp = Vbc.component
-        if comp is None:
-            return BCMixedBitThingy(sub)
-        else:
-            return BCCompOfMixedBitThingy(sub, comp)
+from .bcs import bc2space
 
 
 def getFormDIRK(F, ks, butch, t, dt, u0, bcs=None):
@@ -109,18 +59,16 @@ def getFormDIRK(F, ks, butch, t, dt, u0, bcs=None):
                          dtype=object)
     d_val = MC.Constant(1.0)
     for bc in bcs:
-        Vbc = bc.function_space()
-        dat4bc = getThingy(V, bc)
         bcarg = as_ufl(bc._original_arg)
         bcarg_stage = replace(bcarg, {t: t+c*dt})
 
-        gdat = bcarg_stage - dat4bc(u0)
+        gdat = bcarg_stage - bc2space(bc, u0)
         for i in range(num_stages):
-            gdat -= dt*a_vals[i]*dat4bc(ks[i])
+            gdat -= dt*a_vals[i]*bc2space(bc, ks[i])
 
         gdat /= dt*d_val
 
-        new_bc = DirichletBC(Vbc, gdat, bc.sub_domain)
+        new_bc = bc.reconstruct(g=gdat)
         bcnew.append(new_bc)
 
     return stage_F, (k, g, a, c, a_vals, d_val), bcnew
--- a/irksome/dirk_stepper.py
+++ b/irksome/dirk_stepper.py
@@ -144,6 +144,9 @@ class DIRKTimeStepper:
         self.AAb = numpy.zeros((num_stages+1, num_stages))
         self.AAb[:-1, :] = butcher_tableau.A
         self.AAb[-1, :] = butcher_tableau.b
+        self.CCone = numpy.zeros(num_stages+1)
+        self.CCone[:-1] = butcher_tableau.c
+        self.CCone[-1] = 1.0
 
         self.V = V = u0.function_space()
         self.u0 = u0
@@ -183,13 +186,14 @@ class DIRKTimeStepper:
 
         self.kgac = k, g, a, c, a_vals, d_val
 
-    def update_bc_constants(self, AAb, i, a_vals, d_val):
+    def update_bc_constants(self, AAb, CCone, i, a_vals, d_val, c):
         ns = AAb.shape[1]
         for j in range(i):
             a_vals[j].assign(AAb[i, j])
         for j in range(i, ns):
             a_vals[j].assign(0)
         d_val.assign(AAb[i, i])
+        c.assign(CCone[i])
 
     def advance(self):
         k, g, a, c, a_vals, d_val = self.kgac
@@ -198,14 +202,9 @@ class DIRKTimeStepper:
         dtc = float(self.dt)
         bt = self.butcher_tableau
         AA = bt.A
-        CC = bt.c
         BB = bt.b
         gsplit = g.subfunctions
         for i in range(self.num_stages):
-            # update a, c constants tucked into the variational problem
-            # for the current stage
-            a.assign(AA[i, i])
-            c.assign(CC[i])
             # compute the already-known part of the state in the
             # variational form
             g.assign(u0)
@@ -215,7 +214,8 @@ class DIRKTimeStepper:
                     gbit += dtc * float(AA[i, j]) * kbit
 
             # update BC constants for the variational problem
-            self.update_bc_constants(self.AAb, i, a_vals, d_val)
+            self.update_bc_constants(self.AAb, self.CCone, i, a_vals, d_val, c)
+            a.assign(AA[i, i])
 
             # solve new variational problem, stash the computed
             # stage value.
--- a/irksome/dirk_stepper.py
+++ b/irksome/dirk_stepper.py
@@ -61,6 +61,10 @@ def getFormDIRK(F, ks, butch, t, dt, u0, bcs=None):
     for bc in bcs:
         bcarg = as_ufl(bc._original_arg)
         bcarg_stage = replace(bcarg, {t: t+c*dt})
+        if bcarg_stage == 0:
+            # Homogeneous BC, just zero out stage dofs
+            bcnew.append(bc.reconstruct(g=0))
+            continue
 
         gdat = bcarg_stage - bc2space(bc, u0)
         for i in range(num_stages):
@@ -68,8 +72,7 @@ def getFormDIRK(F, ks, butch, t, dt, u0, bcs=None):
 
         gdat /= dt*d_val
 
-        new_bc = bc.reconstruct(g=gdat)
-        bcnew.append(new_bc)
+        bcnew.append(bc.reconstruct(g=gdat))
 
     return stage_F, (k, g, a, c), bcnew, (a_vals, d_val)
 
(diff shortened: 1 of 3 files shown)


irksome/dirk_stepper.py
```diff
@@ -180,23 +184,27 @@ def __init__(self, F, butcher_tableau, t, dt, u0, bcs=None,
...
+        for j in range(i):
+            a_vals[j].assign(AAb[i, j])
+        for j in range(i, ns):
+            a_vals[j].assign(0)
+        d_val.assign(AAb[i, i])
+        c.assign(CCone[i])
 
     def advance(self):
-        k, g, a, c = self.kgac
+        k, g, a, c, a_vals, d_val = self.kgac
```
pbrubeck, line 199: Please use a better name for this attribute

irksome/dirk_stepper.py
```diff
@@ -101,31 +100,30 @@ def getFormDIRK(F, butch, t, dt, u0, bcs=None):
...
-            gdat = assemble(project(bcarg, Vbc))
-            gmethod = lambda gd, gc: gd.project(gc)
+
+        gdat = bcarg_stage - dat4bc(u0)
+        for i in range(num_stages):
+            gdat -= dt*a_vals[i]*dat4bc(ks[i])
+
+        gdat /= dt*d_val
 
         new_bc = DirichletBC(Vbc, gdat, bc.sub_domain)
```
pbrubeck, line 123:
```suggestion
        new_bc = bc.reconstruct(g=gdat)
```

irksome/dirk_stepper.py
```diff
@@ -101,31 +100,30 @@ def getFormDIRK(F, butch, t, dt, u0, bcs=None):
...
-        try:
-            gdat = assemble(interpolate(bcarg, Vbc))
-            gmethod = lambda gd, gc: gd.interpolate(gc)
-        except:  # noqa: E722
-            gdat = assemble(project(bcarg, Vbc))
-            gmethod = lambda gd, gc: gd.project(gc)
+
+        gdat = bcarg_stage - dat4bc(u0)
+        for i in range(num_stages):
+            gdat -= dt*a_vals[i]*dat4bc(ks[i])
```
pbrubeck, line 119:
```suggestion
            gdat -= dt*a_vals[i]*bc2space(bc, ks[i])
```

irksome/dirk_stepper.py
```diff
@@ -101,31 +100,30 @@ def getFormDIRK(F, butch, t, dt, u0, bcs=None):
...
         bcarg = as_ufl(bc._original_arg)
         bcarg_stage = replace(bcarg, {t: t+c*dt})
-        try:
-            gdat = assemble(interpolate(bcarg, Vbc))
-            gmethod = lambda gd, gc: gd.interpolate(gc)
-        except:  # noqa: E722
-            gdat = assemble(project(bcarg, Vbc))
-            gmethod = lambda gd, gc: gd.project(gc)
+
+        gdat = bcarg_stage - dat4bc(u0)
```
pbrubeck, line 117:
```suggestion
        gdat = bcarg_stage - bc2space(bc, u0)
```

irksome/dirk_stepper.py
```diff
@@ -101,31 +100,30 @@ def getFormDIRK(F, butch, t, dt, u0, bcs=None):
...
 
         new_bc = DirichletBC(Vbc, gdat, bc.sub_domain)
         bcnew.append(new_bc)
 
-        dat4bc = getThingy(V, bc)
-        gdat2 = Function(gdat.function_space())
-        gblah.append((gdat, gdat2, bcarg_stage, gmethod, dat4bc))
-
-    return stage_F, (k, g, a, c), bcnew, gblah
+    return stage_F, (k, g, a, c, a_vals, d_val), bcnew
```
pbrubeck, line 126: Are we returning `(a_vals, d_val)` instead of `gblah`? Is there a good reason to merge these into the original `kgac` tuple?

irksome/dirk_stepper.py
```diff
@@ -142,12 +140,19 @@ def __init__(self, F, butcher_tableau, t, dt, u0, bcs=None,
         self.num_linear_iterations = 0
 
         self.butcher_tableau = butcher_tableau
+        self.num_stages = num_stages = butcher_tableau.num_stages
+        self.AAb = numpy.zeros((num_stages+1, num_stages))
+        self.AAb[:-1, :] = butcher_tableau.A
+        self.AAb[-1, :] = butcher_tableau.b
```
pbrubeck, line 146:
```suggestion
        self.AAb = numpy.vstack((butcher_tableau.A, butcher_tableau.b))
```

irksome/dirk_stepper.py
```diff
@@ -142,12 +140,19 @@ def __init__(self, F, butcher_tableau, t, dt, u0, bcs=None,
...
+        self.CCone = numpy.zeros(num_stages+1)
+        self.CCone[:-1] = butcher_tableau.c
+        self.CCone[-1] = 1.0
```
pbrubeck, line 149:
```suggestion
        self.CCone = numpy.append(butcher_tableau.c, 1.0)
```

irksome/dirk_stepper.py
```diff
@@ -180,23 +184,27 @@ def __init__(self, F, butcher_tableau, t, dt, u0, bcs=None,
             self.problem, appctx=appctx, solver_parameters=solver_parameters,
             nullspace=nullspace)
 
-        self.kgac = k, g, a, c
+        self.kgac = k, g, a, c, a_vals, d_val
+
+    def update_bc_constants(self, AAb, CCone, i, a_vals, d_val, c):
```
pbrubeck, line 189: Almost every argument to this function (except `i`) could be read directly from `self`, so there's no need to pass them as arguments.

irksome/explicit_stepper.py
```diff
@@ -20,3 +20,16 @@ def __init__(self, F, butcher_tableau, t, dt, u0, bcs=None,
             F, butcher_tableau, t, dt, u0, bcs=bcs,
             solver_parameters=solver_parameters, appctx=appctx,
             nullspace=None)
+
+    # DAE treatment of the boundary conditions in this case says that
+    # we should impose the BCs so that they are satisfied for the next
+    # stage for all but that last stage, and that they are satisfied
+    # for the next timestep for the last stage.
+    def update_bc_constants(self, AAb, CCone, i, a_vals, d_val, c):
```
pbrubeck, line 28: Please explain the +1 shift in more detail in the docstring

irksome/explicit_stepper.py
```diff
@@ -20,3 +20,16 @@ def __init__(self, F, butcher_tableau, t, dt, u0, bcs=None,
...
+        ns = AAb.shape[1]
+        for j in range(i):
+            a_vals[j].assign(AAb[i+1, j])
+        for j in range(i, ns):
+            a_vals[j].assign(0)
+        d_val.assign(AAb[i+1, i])
+        c.assign(CCone[i+1])
```
pbrubeck, line 35: Could this shft be avoided by truncating the first row of `AAb`?

irksome/dirk_stepper.py
ScottMacLachlan: Done

irksome/explicit_stepper.py
ScottMacLachlan: I was trying to write this in a way that didn't need any logic to handle the DIRK vs. explicit cases outside of how we update the constants in the BCs.  I guess we could exchange this for having different constructions of `AAb` and `CCone` in the `__init__` class, but I'm not sure we gain anything by it...

irksome/dirk_stepper.py
ScottMacLachlan: My logic was that these are generalizations of the `a` and `c` Constants that we're already returning here, but I agree that the naming is ugly

irksome/dirk_stepper.py
```diff
@@ -180,23 +128,29 @@ def __init__(self, F, butcher_tableau, t, dt, u0, bcs=None,
             self.problem, appctx=appctx, solver_parameters=solver_parameters,
             nullspace=nullspace)
 
-        self.kgac = k, g, a, c
+        self.kgac = k, g, a, c, a_vals, d_val
+
+    def update_bc_constants(self, i, a_vals, d_val, c):
+        AAb = self.AAb
+        CCone = self.CCone
```
pbrubeck, line 135: This should do if `self.butcher_tableau` is a property of the class
```suggestion
        shift = 1 if self.butcher_tableau.is_explicit else 0
        AAb = self.AAb[shift:]
        CCone = self.CCone[shift:]
```

irksome/dirk_stepper.py
ScottMacLachlan: Fixed now

irksome/dirk_stepper.py
ScottMacLachlan: If we do this (and I'd be fine to do it, I'd say that we do this in the `__init__` method, rather than here, since we don't need to do this at every stage of every timestep

irksome/explicit_stepper.py
ScottMacLachlan: I wrote a little more math in the docstring to explain this.  Is that enough?

irksome/dirk_stepper.py
ScottMacLachlan: Now done in the `__init__` method, with comments to document the logic

irksome/dirk_stepper.py
```diff
@@ -101,31 +50,28 @@ def getFormDIRK(F, butch, t, dt, u0, bcs=None):
...
     # than one per stage), but we need a `Function` inside of each BC
     # and a rule for computing that function at each time for each
     # stage.
+    a_vals = numpy.array([MC.Constant(0) for i in range(num_stages)],
+                         dtype=object)
+    d_val = MC.Constant(1.0)
     for bc in bcs:
-        Vbc = bc.function_space()
         bcarg = as_ufl(bc._original_arg)
         bcarg_stage = replace(bcarg, {t: t+c*dt})
```
pbrubeck, line 63:
```suggestion
        bcarg_stage = replace(bcarg, {t: t+c*dt})
        if bcarg_stage == 0:
            # Homogeneous BC, just zero out stage dofs
            bcnew.append(bc)
            continue
```

pbrubeck: We need to treat this as a special case, since we want to allow a scalar 0 data for vector-valued Functions

irksome/dirk_stepper.py
```diff
@@ -101,31 +50,28 @@ def getFormDIRK(F, butch, t, dt, u0, bcs=None):
...
-        gblah.append((gdat, gdat2, bcarg_stage, gmethod, dat4bc))
+        gdat = bcarg_stage - bc2space(bc, u0)
+        for i in range(num_stages):
+            gdat -= dt*a_vals[i]*bc2space(bc, ks[i])
 
-    return stage_F, (k, g, a, c), bcnew, gblah
+        gdat /= dt*d_val
+
+        new_bc = bc.reconstruct(g=gdat)
+        bcnew.append(new_bc)
```
pbrubeck, line 72:
```suggestion
        bcnew.append(bc.reconstruct(g=gdat))
```

irksome/dirk_stepper.py
```diff
@@ -101,31 +50,28 @@ def getFormDIRK(F, butch, t, dt, u0, bcs=None):
...
-    gblah = []
 
     # For the DIRK case, we need one new BC for each old one (rather
     # than one per stage), but we need a `Function` inside of each BC
     # and a rule for computing that function at each time for each
     # stage.
+    a_vals = numpy.array([MC.Constant(0) for i in range(num_stages)],
+                         dtype=object)
+    d_val = MC.Constant(1.0)
     for bc in bcs:
```
pbrubeck, line 61: It'd be nice if we could find a unified treatment for stage BC, and handle everything BC-related within bcs.py. But that could come as a separate PR.

irksome/dirk_stepper.py
```diff
@@ -142,12 +88,33 @@ def __init__(self, F, butcher_tableau, t, dt, u0, bcs=None,
...
+        # For explicit, we say that the stage i solution should be
+        # determined to match the prescribed boundary values at time
+        # c[i+1] (the first stage where it appears), which means we
+        # use the (i+1)^st row of the Butcher tableau in determining
+        # the boundary condition, and the full reconstruction for the
+        # final stage
+
+        shift = 1 if butcher_tableau.is_explicit else 0
+        self.AAb = self.AAb[shift:]
+        self.CCone = self.CCone[shift:]
```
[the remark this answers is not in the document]
ScottMacLachlan: Good call. Done

irksome/dirk_stepper.py
ScottMacLachlan: Done

irksome/dirk_stepper.py
ScottMacLachlan: I was thinking a bit about this as I wrote the code here.  Might not be too difficult, but I agree that it'd be better to save it for a separate PR

irksome/dirk_stepper.py
```diff
@@ -1,67 +1,16 @@
...
 from firedrake import NonlinearVariationalProblem as NLVP
 from firedrake import NonlinearVariationalSolver as NLVS
-from firedrake import assemble, split, project
-from firedrake.__future__ import interpolate
+from firedrake import split
 from ufl.constantvalue import as_ufl
 
 from .deriv import TimeDerivative
 from .tools import replace, MeshConstant
+from .bcs import bc2space
```
rckirby, line 12: Always glad to be rid of thingies.

irksome/dirk_stepper.py
rckirby: I agree that fixing it is good, but a separate issue.  BCS have been the bane of Irksome for most of its existence.  I’m thrilled with the progress y’all are making on this…

irksome/dirk_stepper.py
```diff
@@ -101,31 +50,31 @@ def getFormDIRK(F, butch, t, dt, u0, bcs=None):
...
-        Vbc = bc.function_space()
         bcarg = as_ufl(bc._original_arg)
         bcarg_stage = replace(bcarg, {t: t+c*dt})
-        try:
-            gdat = assemble(interpolate(bcarg, Vbc))
-            gmethod = lambda gd, gc: gd.interpolate(gc)
-        except:  # noqa: E722
-            gdat = assemble(project(bcarg, Vbc))
-            gmethod = lambda gd, gc: gd.project(gc)
+        if bcarg_stage == 0:
```
rckirby, line 64: Do you also want to check for zero of the proper shape in case somebody isn’t taking advantage of just doing zero?

irksome/dirk_stepper.py
```diff
@@ -142,12 +91,33 @@ def __init__(self, F, butcher_tableau, t, dt, u0, bcs=None,
         self.num_linear_iterations = 0
 
         self.butcher_tableau = butcher_tableau
+        self.num_stages = num_stages = butcher_tableau.num_stages
+        self.AAb = numpy.vstack((butcher_tableau.A, butcher_tableau.b))
+        self.CCone = numpy.append(butcher_tableau.c, 1.0)
+
+        # Need to be able to set BCs for either the DIRK or explicit cases.
+
+        # For DIRK, we say that the stage i solution should match the
```
rckirby, line 100: Is it the case that this is sufficiently accurate?  I guess we don’t have an implicit method, so this might be as good as we can get?

irksome/dirk_stepper.py
```diff
@@ -205,21 +182,9 @@ def advance(self):
                 for (gbit, kbit) in zip(gsplit, ksplit):
                     gbit += dtc * float(AA[i, j]) * kbit
 
-            # update BC's for the variational problem
```
rckirby, line 208: Yay for cleaner here.

tests/test_explicit.py
```diff
@@ -0,0 +1,69 @@
...
+import pytest
+from firedrake import *
+from irksome import PEPRK, Dt, MeshConstant, TimeStepper
+from ufl.algorithms.ad import expand_derivatives
+
+peprks = [PEPRK(*x) for x in ((4, 2, 5), (5, 2, 6))]
+
+
+@pytest.mark.parametrize("butcher_tableau", peprks)
+def test_1d_heat_dirichletbc(butcher_tableau):
```
rckirby, line 12: A test is a test, but users do also sometimes read tests for go-by in code usage. Should we be doing a wave equation so as not to give the wrong idea about applicability of explicit methods for the heat equation?

rckirby (approved):
This really looks great.  As long as tests pass, I’m happy with it but do have a few questions.  Not sure if changes to code are needed, or just my understanding.

irksome/dirk_stepper.py
ScottMacLachlan: For the stage values, there's 2 ways to impose BCs via an index-1 DAE, and they both agree that this is the right thing to do.  For DIRK, they would differ in the non-stiffly-accurate casein terms of whether we discard the reconstruction of the BCs that you get from the Butcher Tableau in favour of imposing the BCs at the next timestep.  I guess we could add an option to do that (just in case any non-stiffly-accurate DIRK aficionados are lurking out there), but it didn't seem like a great option to me.

tests/test_explicit.py
ScottMacLachlan: This gets complicated, since the mixed wave equation form wouldn't strongly impose Dirichlet BCs on the solution.  So, I opted to copy a test that we were already using to test BC enforcement (from `test_dirichletbc.py`), and that happened to be heat equation.

I think this boils down to mechanism over policy, in any case

irksome/dirk_stepper.py
ScottMacLachlan: I am happy to do this, but not UFL-savvy enough to know what the proper way to do so is.  I didn't see a similar check anywhere else in Irksome - am I missing something?

tests/test_explicit.py
rckirby: Agreed, one could do the wave equation in H^1 to get strongly enforced BC (or use reflecting BC in mixed form), but this is a test.  Maybe a comment saying “this is only a test of BC feature.  Please don’t actually use explicit methods on the heat equation and then complain to the Irksome developers that our code is busted and gives massive instabilities.

ScottMacLachlan: Warning label added...

rckirby (approved):
Thanks!

PR #104, opened 2024-12-12
Add Galerkin-in-Time

ScottMacLachlan:
We've worked out the UFL manipulation to implement both continuous and discontinuous Galerkin-in-time steppers.  CG is, in some instances, equivalent to Gauss-Legendre (and, so symplectic), while DG is dissipative.


irksome/discontinuous_galerkin_stepper.py
```diff
@@ -0,0 +1,310 @@
...
+                replace(bcarg, {t: t + qpts[q] * dt}))
+        bc_func_for_stages = mmat_inv @ (basis_vals @ bcblah_at_qp)
+        for i in range(num_stages):
+            Vbigi = stage2spaces4bc(bc, V, Vbig, i)
+            bcsnew.append(bc.reconstruct(V=Vbigi, g=bc_func_for_stages[i]))
+
+    return Fnew, UU, bcsnew, getNullspace(V, Vbig, num_stages, nullspace)
+
+
+class DiscGalerkinTimeStepper:
```
pbrubeck, line 188:
```suggestion
class DiscontinuousGalerkinTimeStepper:
```

irksome/discontinuous_galerkin_stepper.py
```diff
@@ -0,0 +1,310 @@
...
+            basis_type = "Lagrange"
+        self.basis_type = basis_type
+
+        V = u0.function_space()
+        self.num_fields = len(V)
+
+        ufc_line = ufc_simplex(1)
+
+        if basis_type == "Lagrange":
+            self.el = DiscontinuousLagrange(ufc_line, order)
```
pbrubeck, line 247: This gives you equispaced points, do you get better stablity/acuracy with a different choice of points? I guess accuracy does not get affected due to basis independence for the quasi-optimality of Galerkin methods.

irksome/discontinuous_galerkin_stepper.py
```diff
@@ -0,0 +1,310 @@
...
+        elif basis_type == "Bernstein":
+            if order == 0:
+                self.el = DiscontinuousLagrange(ufc_line, 0)
+            else:
+                self.el = DiscontinuousElement(
+                    Bernstein(ufc_line, order))
+        else:
+            raise NotImplementedError("Not implemented basis type")
```
pbrubeck, line 255:
```suggestion
        else:
            self.el = DiscontinuousElement(Lagrange(ufc_line, order, variant=basis_type))
```

irksome/discontinuous_galerkin_stepper.py
```diff
@@ -0,0 +1,310 @@
...
+
+        if quadrature is None:
+            quadrature = make_quadrature(ufc_line, order+1)
+        self.quadrature = quadrature
+        assert np.size(quadrature.get_points()) >= order+1
```
pbrubeck, line 260:
```suggestion
        assert np.size(quadrature.get_points()) == order+1
```

irksome/galerkin_stepper.py
```diff
@@ -0,0 +1,295 @@
...
+            self.test_el = DiscontinuousLagrange(ufc_line, order-1)
+        elif basis_type == "Bernstein":
+            self.trial_el = Bernstein(ufc_line, order)
+            if order == 1:
+                self.test_el = DiscontinuousLagrange(ufc_line, 0)
+            else:
+                self.test_el = DiscontinuousElement(
+                    Bernstein(ufc_line, order-1))
+        else:
+            raise NotImplementedError("Not implemented basis type")
```
pbrubeck, line 240:
```suggestion
            self.trial_el = Lagrange(ufc_line, order, variant=basis_type)
            self.test_el = DiscontinuousLagrange(ufc_line, order-1, variant=basis_type)
```

pbrubeck: Similarly, we could add another `basis_type="integral"` with `IntegratedLegendre` and `Legendre`

irksome/discontinuous_galerkin_stepper.py
rckirby: Remember our recent fiat paper -- your have to to *very* high order for this to matter in Galerkin methods for Poisson, I expect similar is true for Galerkin in time.  I didn't realize that `DiscontinuousLagrange` was just equispaced.  Switching to a discontinuous element of Lagrange would fix this.

irksome/galerkin_stepper.py
rckirby: I'm not sure what the use for other Lagrange elements would be, could it possibly diagonalize the mass matrix?  We're in a Petrov-Galerkin setting, this would have to be worked out.

rckirby: If we want to support other kinds of Lagrange basis, maybe `basis_type` could be `"Lagrange"` or `"Lagrange,..."` where we put in a variant after the comma and split it off?

tests/test_disc_galerkin.py
```diff
@@ -0,0 +1,168 @@
...
+import pytest
+from firedrake import *
+from irksome import Dt, MeshConstant, DiscontinuousGalerkinTimeStepper
+from irksome import TimeStepper, RadauIIA
+from ufl.algorithms.ad import expand_derivatives
+import FIAT
+
+
+@pytest.mark.parametrize("order", [0, 1, 2])
+@pytest.mark.parametrize("basis_type", ["Lagrange", ])
```
ScottMacLachlan, line 12: We should be testing more basis types here

irksome/galerkin_stepper.py
ScottMacLachlan: I like how this is resolved now

tests/test_galerkin.py
```diff
@@ -0,0 +1,218 @@
...
+import pytest
+from firedrake import *
+from irksome import Dt, MeshConstant, GalerkinTimeStepper
+from irksome import TimeStepper, GaussLegendre
+from ufl.algorithms.ad import expand_derivatives
+from FIAT import make_quadrature, ufc_simplex
+
+
+@pytest.mark.parametrize("order", [1, 2, 3])
+@pytest.mark.parametrize("basis_type", ["Lagrange", "Bernstein"])
```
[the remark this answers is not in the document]
ScottMacLachlan: I got a local solver error when I added `"integral"` to this test.  Let's see if that persists in CI...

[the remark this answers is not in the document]
ScottMacLachlan: MUMPS fixed this locally for me

irksome/galerkin_stepper.py
```diff
@@ -0,0 +1,274 @@
...
+    tabulate_trials = L_trial.tabulate(1, qpts)
+    trial_vals = tabulate_trials[(0,)]
+    trial_dvals = tabulate_trials[(1,)]
+    test_vals = L_test.tabulate(0, qpts)[(0,)]
+
+    # sort dofs geometrically by entity location
+    edofs = L_trial.entity_dofs()
+    trial_perm = [*edofs[0][0], *edofs[1][0], *edofs[0][1]]
+    trial_vals = trial_vals[trial_perm]
+    trial_dvals = trial_dvals[trial_perm]
```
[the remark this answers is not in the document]
pbrubeck: This should give the right order for point evaluation DOFs, but also handles Bernstein and integral by putting the left/right vertex DOFs as the first and last DOFs.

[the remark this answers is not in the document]
pbrubeck: Test space is DG, the interior dofs should be ordered already.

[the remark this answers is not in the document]
rckirby: FIAT orders the dofs different for continuous and discontinuous Lagrange?

tests/test_galerkin.py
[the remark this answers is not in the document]
rckirby: I can't why regular lu might fail here?  Did you try to twiddle other parameters like shifting?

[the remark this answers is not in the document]
ScottMacLachlan: I did not, but I do wonder if this has to do with a (reasonable) need for pivoting when we use "exotic" bases.  Trying MUMPS is my usually first step in debugging LU failures, so didn't get in to other possible fixes.

irksome/galerkin_stepper.py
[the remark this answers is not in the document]
ScottMacLachlan: I think this is fine, since all DoFs for DG are internal, right?  If so, I guess we're at the mercy of whatever FIAT gives us for their ordering, but probably we shouldn't be second-guessing that without cause

tests/test_disc_galerkin.py
ScottMacLachlan: I've now tested the `basis_type` values that we've implemented.  Are there reasonable choices of `variant` that we should test as well?

rckirby: `integral` since we might be going down that road for coarsening to low-degree in time?

ScottMacLachlan: `integral` is already there, for both DG and CG cases (this is a thread on an outdated commit before I added that)

PR #105, opened 2024-12-17
Fix Constants

pbrubeck:
Some attempts to optimize by introducing tensor-valued `Constants`


rckirby (approved):
Thanks, this gets tests down to 4 minutes from 4.5, probably will help out on more complicated problems down the road!

PR #106, opened 2025-01-10
Add DIRK-IMEX schemes

ScottMacLachlan:
This adds an implementation of the DIRK-IMEX method from the Ascher-Ruuth-Spiteri 1997 Applied Numerical Mathematics paper.  Mostly this layers the IMEX part on top of the existing DIRK code (but is different enough that reuse is not feasible).

### DIFF
--- a/demos/monodomain/demo_monodomain_FHN_dirkimex.py.rst
+++ b/demos/monodomain/demo_monodomain_FHN_dirkimex.py.rst
@@ -145,10 +145,13 @@ Now, we access the IMEX method via the `TimeStepper` as with other methods.  Not
           outfile1.write(uFinal, time=j * float(dt))
           outfile2.write(cFinal, time=j * float(dt))
 
+
+We can print out some solver statistics here.  We expect one implicit solve per stage per timestep, and that's what we see with the four-stage method.  For this Butcher Tableau, we can avoid computing the final explicit stage (since it's coefficient in the next stage reconstruction is zero), so we see the same number of mass solves.::
+
   nsteps, n_nonlin, n_lin, n_nonlin_mass, n_lin_mass = stepper.solver_stats()
   print(f"Time steps taken: {nsteps}")
   print(f"  {n_nonlin} nonlinear steps in implicit stage solves (should be {nsteps*ns})")
   print(f"  {n_lin} linear steps in implicit stage solves")
-  print(f"  {n_nonlin_mass} nonlinear steps in mass solves (should be {nsteps*(ns+1)})")
+  print(f"  {n_nonlin_mass} nonlinear steps in mass solves (should be {nsteps*ns})")
   print(f"  {n_lin_mass} linear steps in mass solves")
 
--- a/demos/monodomain/demo_monodomain_FHN_dirkimex.py.rst
+++ b/demos/monodomain/demo_monodomain_FHN_dirkimex.py.rst
@@ -1,5 +1,5 @@
-Solving monodomain equations with Fitzhugh-Nagumo reaction and a DIRK- IMEX method
-==================================================================================
+Solving monodomain equations with Fitzhugh-Nagumo reaction and a DIRK-IMEX method
+=================================================================================
 
 We're solving monodomain (reaction-diffusion) with a particular reaction term.
 The basic form of the equation is:
@@ -112,7 +112,7 @@ specify only parameters for each stage.  We use an additive Schwarz (fieldsplit)
 	    }}
 
 
-The DIRK-IMEX schemes also require a mass-matrix solver.  Here, we just use AMG on the coupled system, which works fine.::
+The DIRK-IMEX schemes also require a mass-matrix solver.  Here, we just use an incomplete Cholesky preconditioner for CG on the coupled system, which works fine.::
 
   mass_params = {"snes_type": "ksponly",
                  "ksp_rtol": 1.e-8,
--- a/demos/monodomain/demo_monodomain_FHN_dirkimex.py.rst
+++ b/demos/monodomain/demo_monodomain_FHN_dirkimex.py.rst
@@ -1,0 +1,153 @@
+Solving monodomain equations with Fitzhugh-Nagumo reaction and a DIRK- IMEX method
+==================================================================================
+
+We're solving monodomain (reaction-diffusion) with a particular reaction term.
+The basic form of the equation is:
+
+.. math::
+
+   \chi \left( C_m u_t + I_{ion}(u) \right) = \nabla \cdot \sigma \nabla u
+
+where :math:`u` is the membrane potential, :math:`\sigma` is the conductivity tensor, :math:`C_m` is the specific capacitance of the cell membrane, and :math:`\chi` is the surface area to volume ration.  The term :math:`I_{ion}` is current due to ionic flows through channels in the cell membranes, and may couple to a complicated reaction network.  In our case, we take the relatively simple model due to Fitzhugh and Nagumo.  Here, we have a separate concentration variable :math:`c` satisfying the reaction equation:
+
+.. math::
+
+   c_t = \epsilon( u + \beta - \gamma c)
+
+for certain positive parameters :math:`\beta` and :math:`\gamma`, and the current takes the form of:
+
+.. math::
+
+   I_{ion}(u, c) = \tfrac{1}{\epsilon} \left( u - \tfrac{u^3}{3} - c \right)
+
+so that we have an overall system of two equations.  One of them is linear but stiff/diffusive, and the other is nonstiff but nonlinear.  This combination makes the system a good candidate for IMEX-type methods.
+
+
+We start with standard Firedrake/Irksome imports::
+
+  import copy
+
+  from firedrake import (And, Constant, File, Function, FunctionSpace,
+                         RectangleMesh, SpatialCoordinate, TestFunctions,
+                         as_matrix, conditional, dx, grad, inner, split)
+  from irksome import Dt, MeshConstant, IMEX4, TimeStepper
+
+And we set up the mesh and function space.::
+  
+  mesh = RectangleMesh(20, 20, 70, 70, quadrilateral=True)
+  polyOrder = 2
+  
+  V = FunctionSpace(mesh, "CG", 2)
+  Z = V * V
+
+  x, y = SpatialCoordinate(mesh)
+  MC = MeshConstant(mesh)
+  dt = MC.Constant(0.05)
+  t = MC.Constant(0.0)
+
+Specify the physical constants and initial conditions::
+
+  eps = Constant(0.1)
+  beta = Constant(1.0)
+  gamma = Constant(0.5)
+
+  chi = Constant(1.0)
+  capacitance = Constant(1.0)
+
+  sigma1 = sigma2 = 1.0
+  sigma = as_matrix([[sigma1, 0.0], [0.0, sigma2]])
+
+  
+  InitialPotential = conditional(x < 3.5, Constant(2.0), Constant(-1.28791))
+  InitialCell = conditional(And(And(31 <= x, x < 39), And(0 <= y, y < 35)),
+                            Constant(2.0), Constant(-0.5758))
+
+
+  uu = Function(Z)
+  vu, vc = TestFunctions(Z)
+  uu.sub(0).interpolate(InitialPotential)
+  uu.sub(1).interpolate(InitialCell)
+
+  (u, c) = split(uu)
+  
+
+This sets up the Butcher tableau.  All of our IMEX-type methods are
+based on a RadauIIA method for the implicit part.  We use a three-stage method.::
+  
+  butcher_tableau = IMEX4()
+  ns = butcher_tableau.num_stages
+
+To access an IMEX method, we need to separately specify the implicit and explicit parts of the operator.
+The part to be handled implicitly is taken to contain the time derivatives as well::
+  
+  F1 = (inner(chi * capacitance * Dt(u), vu)*dx
+        + inner(grad(u), sigma * grad(vu))*dx
+        + inner(Dt(c), vc)*dx - inner(eps * u, vc)*dx
+        - inner(beta * eps, vc)*dx + inner(gamma * eps * c, vc)*dx)
+
+This is the part to be handled explicitly.::
+	  
+  F2 = - inner((chi/eps) * (-u + (u**3 / 3) + c), vu)*dx
+
+If we wanted to use a fully implicit method, we would just take
+F = F1 - F2.   Note the minus sign, since DIRK-IMEX takes forms as F1 = F2.
+
+Now, set up solver parameters.  Since we're using a DIRK-IMEX scheme, we can
+specify only parameters for each stage::
+  
+  params = {"snes_type": "ksponly",
+            "ksp_monitor": None,
+            "mat_type": "aij",
+            "ksp_type": "fgmres",
+	    "pc_type": "fieldsplit",
+	    "pc_fieldsplit_type": "additive",
+	    "fieldsplit_0": {
+                "ksp_type": "preonly",
+                "pc_type": "gamg",
+	    },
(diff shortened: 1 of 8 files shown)


irksome/imex.py
```diff
@@ -304,3 +307,225 @@ def solver_stats(self):
...
+    k = Function(V)
+    g = Function(V)
+
+    khat = Function(V)
+    ghat = Function(V)
+    vhat = TestFunction(V)
+
+    # If we're on a mixed problem, we need to replace pieces of the
+    # solution.  Stores array of the splittings of the functions for each stage.
+    if num_fields == 1:
```
rckirby, line 332: sometimes these are lists, sometimes `np.array`?  What's up with that?

irksome/imex.py
```diff
@@ -304,3 +307,225 @@ def solver_stats(self):
...
+    Fhat = inner(khat, vhat)*dx - replace(Fexp, replhat)
+
+    bcnew = []
+
+    # For the DIRK-IMEX case, we need one new BC for each old one
+    # (rather than one per stage), but we need a `Function` inside of
+    # each BC and a rule for computing that function at each time for
+    # each stage.
+
+    a_vals = np.array([MC.Constant(0) for i in range(num_stages)],
```
rckirby, line 372: Pablo had done something to get these as a constant array rather than array of constants, which is supposed to improve code gen time.  Check out the galerkin steppers on your branch (which I need to finally merge!)  It doesn't look like this trick is propagated through the RK stuff, so maybe we want to let this ride and come back to RK stuff in a separate PR.

tests/test_imex.py
```diff
@@ -0,0 +1,120 @@
...
+                          (IMEX3(), 3), (IMEX4(), 3)])
+def test_1d_convdiff_neumannbc(butcher_tableau, order):
+    errs = np.array([convdiff_neumannbc(butcher_tableau, order, 10*2**p) for p in [3, 4]])
+    print(errs)
+    conv = np.log2(errs[0]/errs[1])
+    print(conv)
+    assert conv > order-0.4
+
+
+# Note IMEX4 is stiffly accurate, so satisfies BC checks.  IMEX2 and IMEX3 do not
```
rckirby, line 60: bc checks?

rckirby (changes requested):
Thanks, this is lookling really good!  I guess comparisons between DIRKIMEX + FIRK are on the "to-do" list.

demos/monodomain/demo_monodomain_FHN_dirkimex.py.rst
```diff
@@ -0,0 +1,153 @@
...
+  gamma = Constant(0.5)
+
+  chi = Constant(1.0)
+  capacitance = Constant(1.0)
+
+  sigma1 = sigma2 = 1.0
+  sigma = as_matrix([[sigma1, 0.0], [0.0, sigma2]])
+
+  
+  InitialPotential = conditional(x < 3.5, Constant(2.0), Constant(-1.28791))
```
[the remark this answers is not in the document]
ScottMacLachlan: I did steal this from the RadauIIA/IMEX demo, and have fixed in both places

demos/monodomain/demo_monodomain_FHN_dirkimex.py.rst
```diff
@@ -0,0 +1,153 @@
...
+
+  uu = Function(Z)
+  vu, vc = TestFunctions(Z)
+  uu.sub(0).interpolate(InitialPotential)
+  uu.sub(1).interpolate(InitialCell)
+
+  (u, c) = split(uu)
+  
+
+This sets up the Butcher tableau.  All of our IMEX-type methods are
```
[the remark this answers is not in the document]
ScottMacLachlan: Yep - missed this one when I was editing.  Fixed now

demos/monodomain/demo_monodomain_FHN_dirkimex.py.rst
```diff
@@ -0,0 +1,153 @@
...
+
+Now, set up solver parameters.  Since we're using a DIRK-IMEX scheme, we can
+specify only parameters for each stage::
+  
+  params = {"snes_type": "ksponly",
+            "ksp_monitor": None,
+            "mat_type": "aij",
+            "ksp_type": "fgmres",
+	    "pc_type": "fieldsplit",
+	    "pc_fieldsplit_type": "additive",
```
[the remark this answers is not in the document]
ScottMacLachlan: I've documented this more clearly now

demos/monodomain/demo_monodomain_FHN_dirkimex.py.rst
```diff
@@ -0,0 +1,153 @@
...
+	    }}
+
+
+The DIRK-IMEX schemes also require a mass-matrix solver.  Here, we just use AMG on the coupled system, which works fine.::
+
+  mass_params = {"snes_type": "ksponly",
+                 "ksp_rtol": 1.e-8,
+		 "ksp_monitor": None,
+		 "mat_type": "aij",
+		 "ksp_type": "fgmres",
```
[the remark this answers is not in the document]
ScottMacLachlan: This is my standard approach for mass matrices that I don't want to think about - AMG because it "just works", FGMRES so that I don't need to force AMG to be symmetric.  But, CG+ICC works fine here, so no need to use my favourite hammer...

irksome/dirk_imex_tableaux.py
```diff
@@ -0,0 +1,100 @@
...
+        c = np.array([1.0])      # Time steps for implicit part
+        c_hat = np.array([0.0, 1.0])  # Time steps for explicit part
+        order = 1                  # First-order method (Euler)
+        embedded_order = None      # set to None
+        gamma0 = None              # set to None
+        super().__init__(A, b, b_hat, c, order, embedded_order, gamma0)
+        self.A_hat = A_hat
+        self.b_hat = b_hat
+        self.c_hat = c_hat
+        self.is_imex = True  # Mark this as an IMEX scheme
```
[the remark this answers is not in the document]
ScottMacLachlan: The code is assuming we have both DIRK and IMEX, so this is used just to check that.  But your point is well-taken, that this is badly named.  I've renamed the property to `is_dirk_imex`

irksome/imex.py
ScottMacLachlan: Good question (and one that I asked myself).  This is copied/extended from [here](https://github.com/firedrakeproject/Irksome/blob/cbf6af426277fd84369c3fa6cdea211305d004be/irksome/dirk_stepper.py#L29).  It ain't broke, but I'm not opposed to fixing it.

tests/test_imex.py
ScottMacLachlan: In this demo, we explicitly check that the BCs are satisfied at each timestep.  With DAE-style BCs, that's true only for the stiffly accurate schemes.  I've made the comment less cryptic

irksome/dirk_imex_tableaux.py
```diff
@@ -0,0 +1,100 @@
+from .ButcherTableaux import ButcherTableau
+import numpy as np
+
+
+# IMEX Butcher tableau using numpy arrays
+class IMEXEuler(ButcherTableau):
```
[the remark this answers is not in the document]
ScottMacLachlan: Ascher-Ruuth-Spiteri have a numbering system that obfuscates the schemes to me, writing `(s, sigma, p)`, where `s` is the number of stages, `sigma` is the number of explicit stages needed to be calculated (sometimes `s`, sometimes `s+1` -- we currently assume it is always `s+1`, which means we sometimes do an unnecessary mass solve), and `p` is order of the scheme.  I'd still prefer to avoid that, so perhaps just the factory you describe is sufficient naming?

demos/monodomain/demo_monodomain_FHN_dirkimex.py.rst
```diff
@@ -0,0 +1,153 @@
...
+  uu.sub(0).interpolate(InitialPotential)
+  uu.sub(1).interpolate(InitialCell)
+
+  (u, c) = split(uu)
+  
+
+This sets up the Butcher tableau.  All of our IMEX-type methods are
+based on a RadauIIA method for the implicit part.  We use a three-stage method.::
+  
+  butcher_tableau = IMEX4()
```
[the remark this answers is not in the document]
ScottMacLachlan: Let's discuss/fix this there

irksome/imex.py
ScottMacLachlan: I saw that change on the Galerkin branch, but wasn't sure.  I agree - let's fix that everywhere in a separate PR.

rckirby: What is the purpose of an array of `num_stages`  zeros (wrapped as constant), anyway.  Do these get updated somewhere?

rckirby: And the key idea is putting a numpy array inside of `Constant` rather than vectorizing `Constant`.  It's pretty simple.

irksome/stepper.py
```diff
@@ -161,6 +162,17 @@ def TimeStepper(F, butcher_tableau, t, dt, u0, **kwargs):
             it_solver_parameters, prop_solver_parameters,
             splitting, appctx, nullspace,
             num_its_initial, num_its_per_step)
+    elif stage_type == "dirkimex":
+        Fexp = kwargs.get("Fexp")
+        assert Fexp is not None
```
rckirby, line 167: Or, issue a warning that we don't have an explicit part and are defaulting to an IMEX method.  At ay rate, the failure/warning
 should have an error message to this effect.

rckirby (changes requested):
Thanks for the updates.  Not sure what to do yet about lacking the optimization from Spiteri's paper.

irksome/dirk_imex_tableaux.py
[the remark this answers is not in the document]
rckirby: Maybe the right thing to do is write out the loop "both ways" (the general case and optimized to avoid the extra mass solve).  Maybe there is a clever way to fuse the loops, or maybe we just want an "if/else" branching on whether that optimization is possible.  We could create `advance_normal` and `advance_optimized` methods and set `self.advance` to be one of these in the constructor.  It would look like code duplication, but avoiding the extra mass solve is worthwhile if you're trying to use IMEX as a performance win.

irksome/imex.py
rckirby: N/M, I see the updates.  They get re-set in the loop inside of `advance`.  Also, this is just a vector, not a matrix, so may not be a major cause for alarm.

irksome/dirk_imex_tableaux.py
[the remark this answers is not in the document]
ScottMacLachlan: What about a small refactor to loop over the solves as each stages' mass solve then implicit solve for the first `ns` stages, then having `finalize_normal` (compute last explicit stage and general-case update), `finalize_no_last_explicit` (general-case update where last explicit stage is not used), and `finalize_stiffly_accurate` (update where last explicit stage is not used, and last implicit stage is new `u0`)?  That would get us out of code duplication altogether, and we could sniff which case in `__init__`

irksome/stepper.py
ScottMacLachlan: This was copied from [here](https://github.com/firedrakeproject/Irksome/blob/cbf6af426277fd84369c3fa6cdea211305d004be/irksome/stepper.py#L149), but I agree that a warning is a good idea.  Will add that now in both places.

irksome/dirk_imex_tableaux.py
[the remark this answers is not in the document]
rckirby: I think that's a great plan.  Does it get the optimal pattern for each case?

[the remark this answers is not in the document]
ScottMacLachlan: One way to find out...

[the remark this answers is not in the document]
ScottMacLachlan: Okay - 6a9a568 passes tests locally with the last mass solve and the final update separated into `_final_general`.  Now to implement the special cases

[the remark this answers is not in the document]
ScottMacLachlan: And 02a8441 implements the special cases.  So, seems like this is good to me.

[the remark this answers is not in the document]
ScottMacLachlan: Which brings us back to naming...  Shall I just implement a factory that uses the Ascher-Ruuth-Spiteri indexing convention?  Seems the easiest for users, even if it is a bit opaque

irksome/imex.py
```diff
@@ -506,11 +519,27 @@ def advance(self):
             for ghatbit, kbit in zip(ghat.subfunctions, ks[i].subfunctions):
                 ghatbit += dtc * AA[i, i] * kbit
 
-            chat.assign(C_hat[i+1])
-            self.mass_solver.solve()
-            self.num_mass_nonlinear_iterations += self.mass_solver.snes.getIterationNumber()
-            self.num_mass_linear_iterations += self.mass_solver.snes.getLinearSolveIterations()
-            k_hat_s[i + 1].assign(khat)
+        self._finalize()
```
rckirby, line 522: This seems reasonable.  Do the four IMEX/DIRK tableaux you've implemented actually hit all of these cases in the test suite?

irksome/dirk_imex_tableaux.py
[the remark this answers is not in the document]
rckirby: I think the factory approach is probably cleaner for the user interface.

irksome/imex.py
ScottMacLachlan: I could go fishing for more methods, but I don't actually know of one that would allow `_finalize_no_last_explicit` and doesn't satisfy the conditions for `_finalize_stiffly_accurate`.  None of the ones in the Ascher-Ruuth-Spiteri paper would do so.

I tested `_finalize_no_last_explicit` locally, just by hacking the code to not call the other routine.  Another option would just be a specialized test that overrides `self._finalize` after calling `__init__` in order to be sure we hit that code.

irksome/dirk_imex_tableaux.py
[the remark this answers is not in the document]
ScottMacLachlan: Factory is now implemented, in a2d1dd8

irksome/imex.py
rckirby: If there aren't any methods that actually go that route, we're probably fine.

ScottMacLachlan: Yes - presumably if we add new methods that exercise the code, we'll also add new tests...

demos/monodomain/demo_monodomain_FHN_dirkimex.py.rst
```diff
@@ -0,0 +1,157 @@
...
+The part to be handled implicitly is taken to contain the time derivatives as well::
+  
+  F1 = (inner(chi * capacitance * Dt(u), vu)*dx
+        + inner(grad(u), sigma * grad(vu))*dx
+        + inner(Dt(c), vc)*dx - inner(eps * u, vc)*dx
+        - inner(beta * eps, vc)*dx + inner(gamma * eps * c, vc)*dx)
+
+This is the part to be handled explicitly.::
+	  
+  F2 = - inner((chi/eps) * (-u + (u**3 / 3) + c), vu)*dx
```
rckirby, line 91: Please confirm that signs are consistent between RadauIIA-IMEX Buvoli and DIRK-IMEX?  That would be a painful reason for users to get the wrong answer...

rckirby (approved):
Up to making sure that our two flavors of IMEX are consistent, looks good to me!

demos/monodomain/demo_monodomain_FHN_dirkimex.py.rst
ScottMacLachlan: Signs were inconsistent here, since the code was consistent with signs from the Ascher-Ruuth-Spiteri paper.  I've reversed the signs in 36504a9, so they are now consistent with what is in the RadauIIA-IMEX case, and documented to make this clear

PR #112, opened 2025-02-10
New base class DIRK_IMEX

jbigmanUT:
Created new base class DIRK_IMEX which now ARS_DIRK_IMEX inherits from. Implicit tableau in ARS methods are now padded so sizing is uniform with methods requiring an implicit first step.

Modified DIRKIMEXMethod to allow for implicit first stages. Also combined loops for efficiency

Added remaining ARS schemes for completeness

Added some SSPk schemes from Pareschi and Russo to demonstrate new functionality in SSPK_DIRK_IMEX

Updated tests and demos accordingly


irksome/__init__.py
```diff
@@ -8,7 +8,7 @@
 from .ButcherTableaux import RadauIIA       # noqa: F401
 from .pep_explicit_rk import PEPRK          # noqa: F401
 from .deriv import Dt                       # noqa: F401
-from .dirk_imex_tableaux import DIRK_IMEX   # noqa: F401
+from .dirk_imex_tableaux import DIRK_IMEX, ARS_DIRK_IMEX, SSPK_DIRK_IMEX   # noqa: F401
```
ScottMacLachlan, line 11: Our style here seems to be one import per line

tests/test_imex.py
```diff
@@ -126,6 +133,16 @@ def test_1d_heat_dirichletbc(imp_stages, exp_stages, order):
         assert isclose(u.at(x1), u_1)
 
 
+# Note that ARS_DIRK_IMEX(1,1,1), ARS_DIRK_IMEX(2, 2, 2), and ARS_DIRK_IMEX(4,4,2)
```
ScottMacLachlan, line 136: It's not your typo, but this should be ARS_DIRK_IMEX(4,4,3), shouldn't it?

ScottMacLachlan (commented):
Overall, this looks good to me - many thanks for the great work here!  A couple of small issues to fix, and it would be great to get @rckirby's eyes on this as well.

I think the correct fix for the docs build is to edit line 51 of `.github/workflows/docs.yml` to take v2.5.0 to v3

irksome/__init__.py
jbigmanUT: I separated this into three lines on the latest commit

tests/test_imex.py
jbigmanUT: Good catch, I updated that as well

irksome/dirk_imex_tableaux.py
```diff
@@ -1,71 +1,97 @@
...
+
+        # Expand A, b, c with assumed zeros in ARS tableaux
+        A = self._pad_matrix(A, "lr")
+        b = np.append(np.zeros(1), b)
+        c = np.append(np.zeros(1), c)
+
+        super(ARS_DIRK_IMEX, self).__init__(A, b, c, A_hat, b_hat, c_hat, order)
+
+
+class SSPK_DIRK_IMEX(DIRK_IMEX):
```
rckirby, line 84: I suggest moving this into the `sspk_dirk_imex_tableau.py` file (see below).  Also, you have everything you need right here to provide a comparable factory for `SSPRK_Butcher_Tableaux` for the the standard SSP files (just copy paste + simplify), and that's a useful extra contribution if you care to make it.

rckirby (changes requested):
Thanks, this looks really good.  I've made a few suggestions but think this is in great shape.  Also, I approved the CI to start running (I had to do that since it's from a fork), so we want to make sure that goes through.

rckirby:
@jbigmanUT : @ScottMacLachlan just fixed the doc building issue for us, and it's merged.  Can you merge our master branch to fix that on your PR, as we don't have write permission for you.

tests/test_imex.py
```diff
@@ -47,24 +47,31 @@ def convdiff_neumannbc(butcher_tableau, order, N):
...
                           (3, 4, 3), (4, 4, 3)])
-def test_1d_convdiff_neumannbc(imp_stages, exp_stages, order):
-    bt = DIRK_IMEX(imp_stages, exp_stages, order)
+def test_1d_ars_convdiff_neumannbc(imp_stages, exp_stages, order):
+    bt = ARS_DIRK_IMEX(imp_stages, exp_stages, order)
     errs = np.array([convdiff_neumannbc(bt, order, 10*2**p) for p in [3, 4]])
     print(errs)
     conv = np.log2(errs[0]/errs[1])
     print(conv)
     assert conv > order-0.4
```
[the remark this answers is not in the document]
jbigmanUT: Done, the parameterization is now over the tableaux. I added `ids` for readability

irksome/sspk_dirk_imex_tableau.py
```diff
@@ -0,0 +1,41 @@
+import numpy as np
+
+# Butcher tableau based on schemes in Pareschi and Russo Journal of Scientific Computing 2005
+
```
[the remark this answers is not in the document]
jbigmanUT: I moved the dictionaries for ARS and SSPk methods into the same file as their classes

irksome/sspk_dirk_imex_tableau.py
```diff
@@ -0,0 +1,41 @@
+import numpy as np
```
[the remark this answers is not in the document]
jbigmanUT: I took a look at that GST paper. Since they're written as a recursive scheme instead of via tableau, it wasn't immediately obvious to me they were the same. I'm not doubting it, just not familiar enough to confirm this without extensive effort.

That said, I separated the explicit parts into `SSPButcherTableau` and added them to the testing for explicit schemes in `test_explicit.py`

ScottMacLachlan (approved):
This looks good to me.  Thanks!

rckirby (approved):
Many thanks, I appreciate the contribution!

PR #114, opened 2025-02-12
Fix doc build

ScottMacLachlan:
We're seeing unrelated failures in the doc build in #112, that are difficult to debug since that's from a fork.  Trying to fix these here, for efficiency.


rckirby (approved):
Thanks for the fix!

PR #118, opened 2025-02-20
Refactor replacements

pbrubeck:
Executive summary:

- `component_replace` takes a dictionary of replacements and programmatically inserts replacements of all sub-components.  This was a commonly used pattern, so tucking it into a function simplifies things.
- it is also easier to iterate over scalar components of a ufl `Function` or `TestFunction` by indexing, so we don't have to use `split` and then worry whether the pieces were scalar or vector-valued.  This is largely hidden inside `component_replace`
- `TimeDerivative` now pushes through tensors during construction, and this greatly simplified some of the replacements.

These key changes required pervasive simplification of all of our the UFL manipulation in our various time steppers.

We explored `Constant` around numpy versus numpy array of `Constant` objects.  The former makes fewer GEM nodes so might help with compilation, but the latter seems needed most of the time to respect sparsity for computation.

### DIFF
--- a/irksome/imex.py
+++ b/irksome/imex.py
@@ -9,7 +9,7 @@ from ufl.classes import Zero
 from .ButcherTableaux import RadauIIA
 from .deriv import TimeDerivative
 from .stage import getFormStage
-from .tools import AI, ConstantOrZero, IA, MeshConstant, replace
+from .tools import AI, ConstantOrZero, IA, MeshConstant, component_replace
 from .bcs import bc2space
 
 
@@ -49,7 +49,7 @@ def getFormExplicit(Fexp, butch, u0, UU, t, dt, splitting=None):
     Aexp = riia_explicit_coeffs(num_stages)
 
     vecconst = np.vectorize(ConstantOrZero)
-    
+
     Aprop = vecconst(Aexp)
     Ait = vecconst(butch.A)
     C = vecconst(butch.c)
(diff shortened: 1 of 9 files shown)


irksome/stage.py
```diff
@@ -118,18 +92,18 @@ def getFormStage(F, butch, u0, t, dt, bcs=None, splitting=None, vandermonde=None
     ZZ = Function(Vbig)
 
     # set up the pieces we need to work with to do our substitutions
-    u0bits, vbits, VVbits, ZZbits = getBits(num_stages, num_fields,
-                                            u0, ZZ, v, VV)
+    # u0bits, vbits, VVbits, ZZbits = getBits(num_stages, num_fields,
+    #                                         u0, ZZ, v, VV)
+
```
pbrubeck, line 97:
```suggestion
```

tests/test_dirk.py
```diff
@@ -257,7 +257,7 @@ def test_stokes_bcs(butcher_tableau, bctype):
          - inner(q, div(u))*dx
          - inner(u_rhs, v)*dx
          - inner(p_rhs, q)*dx)
-    Fdirk = replace(F, {u: u_dirk, p: p_dirk})
+    Fdirk = replace(F, {z[i]: z_dirk[i] for i in numpy.ndindex(z.ufl_shape)})
```
pbrubeck, line 260:
```suggestion
    Fdirk = replace(F, {z: z_dirk})
```

irksome/imex.py
```diff
@@ -251,6 +226,7 @@ def __init__(self, F, Fexp, butcher_tableau,
 
         num_fields = len(self.u0.function_space())
         u0split = u0.subfunctions
+
```
pbrubeck, line 229:
```suggestion
```

tests/test_dirk.py
```diff
@@ -1,5 +1,5 @@
 from math import isclose
-
+import numpy
```
pbrubeck, line 2:
```suggestion
```

ScottMacLachlan (approved):
This all looks great to me - many thanks for taking this on!

PR #121, opened 2025-02-25
Mixed FunctionSpaces in DIRKIMEX

jmbigman:


pbrubeck:
Could you also include a test? Thank you for contributing!

ScottMacLachlan:
Ah - this explains a lot... Since the existing tests on the DIRK-IMEX code all use a single equation, we wouldn't have caught this.

I agree with @pbrubeck that this suggests a test.  Maybe the natural thing to do is a simple 2D convection-diffusion equation written as a system (in the style of mixed Poisson), with the convection treated explicitly?

ScottMacLachlan:
@jmbigman : Can you please merge master again, and then we should have happy CI... Thanks!

PR #123, opened 2025-02-26
Add pyproject.toml and adapt to new containers

connorjward:
I don't think that these will pass until the Firedrake containers are rebuilt


rckirby (approved):
Thanks for handling this.  I'm going to assume you know what you're doing, and the lights are green...

rckirby (changes requested):
This is failing.  Not sure why?

pyproject.toml
```diff
@@ -0,0 +1,27 @@
+[build-system]
+requires = ["setuptools"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "IRKsome"
+version = "0.0.1"
```
ScottMacLachlan, line 7: I know it doesn't really matter, but feels like we might be justified to bump this version number...

pyproject.toml
```diff
@@ -0,0 +1,27 @@
...
+authors = [
+  {name = "Rob Kirby", email = "Robert_Kirby@baylor.edu"},
+  {name = "Jorge Marchena Menendez"},
```
ScottMacLachlan, line 10: We updated an author list when we fixed the docs a couple of months ago.  Shall we use that here?

connorjward: Sure I can do that. I just used what was in the setup.py. Alternatively you could omit authors here and set a maintainers field instead

rckirby: What does that look like?  just replacing `authors` with `maintainers` (and appropriate details inside)?

pyproject.toml
rckirby: Do you have a suggestion as to what number we're on by now?

connorjward: If you're not committing to making versioned releases (almost certainly overkill) then I don't think this matters at all.

pyproject.toml
connorjward: Yes. It isn't really important though. I will use the authors list defined in `conf.py`.

pyproject.toml
ScottMacLachlan: Fair enough

rckirby (approved):
If Connor is happy with it, I'm glad to go.

PR #126, opened 2025-02-28
Try to temp fix CI

ScottMacLachlan:
Try to point to older firedrake containers, in the hopes that CI runs


connorjward (approved):
Looks good to me. I'll revisit with the updated containers when the bugs are ironed out.

rckirby (approved):
Fine with me.

PR #128, opened 2025-03-04
Enhacements for TimeDerivative

pbrubeck:
This PR adds support for `Dt(expr(t), order)` where `expr(t)` can be a general time-dependent UFL expression, and `order` >= 0 can be arbitrary.


rckirby (commented):
Do we need a test for this somehow?

PR #129, opened 2025-03-04
Nystrom

rckirby:
Provides support for a basic Runge-Kutta-Nystrom method (stage derivative, no reformulation) for 2nd-order in time.  A simple wave equation test passes.


irksome/nystrom_stepper.py
```diff
@@ -0,0 +1,137 @@
+from .base_time_stepper import StageCoupledTimeStepper
+from .bcs import BCStageData, bc2space
+from .deriv import TimeDerivative
+from .tools import component_replace, replace, vecconst
+from firedrake import TestFunction, as_ufl
+import numpy
+from ufl import zero
+
+
+class NystromTableau:
```
pbrubeck, line 10: Should ButcherTableau inherit from this class?

irksome/nystrom_stepper.py
```diff
@@ -0,0 +1,137 @@
...
+                      + sum(kp[nf * s + i] * (bbar[s] * dt**2)
+                            for s in range(ns)))
+            ut0bit += sum(kp[nf * s + i] * (b[s] * dt) for s in range(ns))
+
+    def get_form_and_bcs(self, stages, tableau=None):
+        if tableau is None:
+            tableau = self.tableau
+        return getFormNystrom(self.F, tableau, self.t,
+                              self.dt, self.u0, self.ut0,
+                              self.stages,
```
pbrubeck, line 135:
```suggestion
                              stages,
```

irksome/nystrom_stepper.py
rckirby: Maybe the other way (this inherits from BT)?  Not sure of the optimal design here.  Every Nystrom tableau has A, b, c but not every Butcher tableau has Abar, bbar .  Every regular Butcher tableau can be promoted by `Abar = A @ A` and `bbar = A.T @ b`.  However, supporting queries (is this method explicit/implicit etc?) has to be overloaded in the Nystrom case since you have two matrices to check instead of one.

irksome/nystrom_stepper.py
rckirby: Good catch.

irksome/nystrom_stepper.py
```diff
@@ -0,0 +1,137 @@
...
+    dtu = TimeDerivative(u0)
+    dt2u = TimeDerivative(dtu)
+
+    Fnew = zero()
+
+    for i in range(num_stages):
+        repl = {t: t + c[i] * dt,
+                v: v_np[i],
+                u0: u0 + c[i] * dt * ut0 + dt**2 * Abark[i],
+                dtu: ut0 + dt * Ak[i],
```
pbrubeck, line 61:
```suggestion
                u0: u0 + ut0 * (c[i] * dt) + Abark[i] * dt**2,
                dtu: ut0 + Ak[i] * dt,
```

rckirby (commented):
Some more things are needed.  At least, we need

- To test something besides 1d wave equation
- A demo?

irksome/nystrom_stepper.py
ScottMacLachlan: I'm with @rckirby here - I think Nystrom inherits from Butcher, since it's Butcher + more stuff.  

I don't know that there's much to be done about it, but I notice that we're starting to propagate slightly ambiguous additional data in these Tableau.  Nystrom has `Abar` and `bbar`, DIRK-IMEX have `A_hat` and `b_hat`.  Regular Butcher Tableau have `btilde` for possibly adaptivity.  It's not our fault, but collisions here appear inevitable.  For example, the notation in Hairer-Wanner-Lubich for Partitioned schemes have `A_hat` and `b_hat` again.  Do we care?  Is there a good way to keep track of these?

irksome/nystrom_stepper.py
```diff
@@ -0,0 +1,137 @@
...
+
+    @property
+    def num_stages(self):
+        return len(self.b)
+
+
+def butcher_to_nystrom(butch):
+    A = butch.A
+    b = butch.b
+    return NystromTableau(A, b, butch.c, A @ A, A.T @ b, butch.order)
```
ScottMacLachlan, line 27: Do we want a `partitioned_to_nystrom` that takes 2 BTs and spits out the Nystrom Tableau that results?

ScottMacLachlan (commented):
Looks good so far.  I agree with @rckirby that this needs a demo and some more tests. Only a few other small things that come to mind reading the code...

irksome/nystrom_stepper.py
rckirby: Structurally, we have common fields between Nystrom & partitioned, but they do different things.  I think there is a high potential for confusion if we try to make them share code?

Inheriting Nystrom from standard Butcher is mathematically a bit dubious -- there exist Nystrom methods that aren't derived from RK methods.  If a Nystrom tableau IS a Butcher tableau (in the sense of inheritance), that should guarantee one could use it for a 1st order problem.  Which isn't true.

irksome/nystrom_stepper.py
rckirby: Partitioned methods are for systems already written in 1st order form, just split into two pieces (and so applicable in principle to second order problems).  If we had partitioned methods, we could (and probably should) do this conversion, and maybe have the stepper automatically call the conversion for us.

demos/nystrom/demo_wave.py.rst
```diff
@@ -0,0 +1,109 @@
...
+space :math:`V`.
+
+We transform this into weak form by multiplying by an arbitrary test function
+:math:`v\in V` and integrating over :math:`\Omega`.  We know have the
+variational problem of finding :math:`u:[0,T]\rightarrow V` such
+that
+
+.. math::
+
+   (u_t, v) + (\nabla u, \nabla v) = (f, v)
```
pbrubeck, line 24:
```suggestion
   (u_{tt}, v) + (\nabla u, \nabla v) = (f, v)
```

demos/nystrom/demo_wave.py.rst
```diff
@@ -0,0 +1,109 @@
...
+  uinit = sin(pi * x) * cos(pi * y)
+  u = Function(V)
+  u.interpolate(uinit)
+  ut = Function(V)
+
+Now, we will define the semidiscrete variational problem using
+standard UFL notation, augmented by the ``Dt`` operator from Irksome::
+
+  v = TestFunction(V)
+  F = inner(Dt(u, 2), v)*dx + inner(grad(u), grad(v))*dx
```
pbrubeck, line 76: We could add some explanation on the second argument passed to `Dt`.

demos/nystrom/demo_wave.py.rst
```diff
@@ -0,0 +1,109 @@
...
+stages at each time step.::
+
+  stepper = StageDerivativeNystromTimeStepper(
+      F, butcher_tableau, t, dt, u, ut, bcs=bc,
+      solver_parameters=luparams)
+
+The system energy is an important quantity for the wave equation, and it
+should be conserved with our choice of time-stepping method::
+
+  E = 0.5 * (inner(ut, ut) * dx + inner(grad(u), grad(u)) * dx)
```
pbrubeck, line 97: `ut` is never updated, so this energy functional is only valid at the initial condition.

rckirby:
And what happens if we have some fields with two time derivatives and others with just one?  There are some such applications, and it would be good to have that support.

irksome/nystrom_stepper.py
ScottMacLachlan: I'm convinced that we should have constructors for Butcher -> Nystrom and (eventually) partitioned -> Nystrom, but not inheritance.

That doesn't remove the need to discuss our naming convention problem as we graft additional data on to a standard Butcher Tableau.  We can push that down the road (and I'm happy to do so), but we are starting to run into ambiguity with our `A_hat`, `Abar`, etc., naming.

irksome/nystrom_stepper.py
ScottMacLachlan: I am thinking about schemes like Störmer-Verlet and other Lobatto IIIA-IIIB pairs, where there is an underlying partitioned scheme (that we have yet to implement), that leads to a sensible Nyström method.  I'd rather be able to construct those programatically (admittedly we don't have Lobatto IIIB implemented yet) than in an ad-hoc way.  I'm happy to push that constructor down the road, though, until it is more relevant.

rckirby: Agreed these are good and worth doing (if we can figure out efficiency) and once we do, the tableaux can programmatically be munged into Nystrom.  I think we should wait till we tackle those later

ScottMacLachlan (commented):
BCs definitely need a fix, otherwise this is mostly small stuff (lots of fixes in the demos, in particular).  Thanks!

demos/nystrom/demo_wave.py.rst
```diff
@@ -0,0 +1,107 @@
+Solving the Heat Equation with Irksome
+======================================
+
+Let's start with the simple wave equation on :math:`\Omega = [0,1]
+\times [0,1]`, with boundary :math:`\Gamma`:
+
+.. math::
+
+   u_{tt} - \Delta u &= f
```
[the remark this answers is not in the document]
rckirby: Agreed.

demos/nystrom/demo_wave.py.rst
```diff
@@ -0,0 +1,107 @@
...
+Let's use simple solver options::
+
+  luparams = {"mat_type": "aij",
+              "ksp_type": "preonly",
+              "pc_type": "lu"}
+
+Most of Irksome's magic happens in the :class:`.TimeStepper`.  It
+transforms our semidiscrete form `F` into a fully discrete form for
+the stage unknowns and sets up a variational problem to solve for the
+stages at each time step.::
```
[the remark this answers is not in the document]
rckirby: Yes

demos/nystrom/demo_wave_mg.py.rst
```diff
@@ -0,0 +1,90 @@
...
+within Firedrake::
+
+  N = 4
+  nref = 3
+  base = UnitSquareMesh(N, N)
+  mh = MeshHierarchy(base, nref)
+  msh = mh[-1]
+
+From here, setting up the function space, manufactured solution, etc,
+are just as for the regular heat equation demo::
```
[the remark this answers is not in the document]
rckirby: Yep

demos/nystrom/demo_wave_mg.py.rst
```diff
@@ -0,0 +1,90 @@
...
+  ut = Function(V)
+
+  v = TestFunction(V)
+
+  F = inner(Dt(u, 2), v)*dx + inner(grad(u), grad(v))*dx
+  bc = DirichletBC(V, 0, "on_boundary")
+
+And now for the solver parameters.  Note that we are solving a
+block-wise system with all stages coupled together.  This performs a
+monolithic multigrid with pointwise block Jacobi preconditioning::
```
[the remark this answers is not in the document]
rckirby: Commentary is updated.

irksome/nystrom_stepper.py
```diff
@@ -0,0 +1,172 @@
...
+        try:
+            bA1inv = numpy.linalg.inv(tableau.A)
+            A1inv = vecconst(bA1inv)
+        except numpy.linalg.LinAlgError:
+            raise NotImplementedError("Can't have DAE BC's for this method")
+
+        def bc2gcur(bc, i):
+            gorig = as_ufl(bc._original_arg)
+            ucur = bc2space(bc, u0)
+            gcur = (1/dt) * sum((replace(gorig, {t: t + c[j]*dt}) - ucur) * A1inv[i, j]
```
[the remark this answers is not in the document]
rckirby: Fixed now.  Thanks for the input

ScottMacLachlan (approved):
Looks good.  We have two tests of Nystrom, both 1D scalar problems, plus the two 2D demos.  The demos have homogeneous Dirichlet BCs, one test has non homogeneous Dirichlet BCs, one has no BCs.  Are there other cases we should be looking for coverage of?

Is there a meaningful mixed FEM example to consider in this setting, just to test things on product spaces?

tests/test_differentiation.py
```diff
@@ -0,0 +1,68 @@
...
+    u = Function(V)
+    w = Function(V)
+    expr = Dt(Dt(dot(u, w)))
+
+    expr = expand_time_derivatives(expr)
+    expected = (dot(Dt(u, 2), w)
+                + dot(Dt(u), Dt(w))
+                + dot(Dt(u), Dt(w))
+                + dot(u, Dt(w, 2)))
+    # UFL equality is failing here due to different index numbers
```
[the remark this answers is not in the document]
rckirby: I made the comment less cryptic.

irksome/stage_derivative.py
```diff
@@ -77,6 +80,7 @@ def getForm(F, butch, t, dt, u0, stages, bcs=None, bc_type=None, splitting=AI):
     # preprocess time derivatives
     F = expand_time_derivatives(F, t=t, timedep_coeffs=(u0,))
     dtu = TimeDerivative(u0)
+    F = expand_time_derivatives(F, t=t, timedep_coeffs=(u0,))
```
pbrubeck, line 83:
```suggestion
```

irksome/stage_derivative.py
```diff
@@ -77,6 +80,7 @@ def getForm(F, butch, t, dt, u0, stages, bcs=None, bc_type=None, splitting=AI):
     # preprocess time derivatives
     F = expand_time_derivatives(F, t=t, timedep_coeffs=(u0,))
```
pbrubeck, line 81:
```suggestion
```

demos/nystrom/demo_wave.py.rst
```diff
@@ -0,0 +1,103 @@
...
+
+  from irksome import GaussLegendre, Dt, MeshConstant, StageDerivativeNystromTimeStepper
+
+We will create the Butcher tableau for a Gauss-Legendre
+Runge-Kutta method, which generalizes the implicit midpoint rule::
+
+  ns = 2
+  butcher_tableau = GaussLegendre(ns)
+
+Now we define the mesh and piecewise linear approximating space in
```
ScottMacLachlan, line 40: `linear` -> `quadratic`

irksome/nystrom_stepper.py
```diff
@@ -0,0 +1,190 @@
...
+        self.bbar = bbar
+        self.order = order
+
+    @property
+    def num_stages(self):
+        return len(self.b)
+
+    @property
+    def is_explicit(self):
+        return (numpy.allclose(self.Abar, 0)
```
ScottMacLachlan, line 25: Should be `numpy.allclose(numpy.triu(self.Abar), 0)` here, too

ScottMacLachlan (approved):
Couple of small things; otherwise looks good

irksome/nystrom_stepper.py
rckirby: Got it

demos/nystrom/demo_wave.py.rst
rckirby: Yes.

rckirby:
> Is there a meaningful mixed FEM example to consider in this setting, just to test things on product spaces?

I've got one started.  Didn't get it pushed (mixed wave `p_{tt} + div(u) = 0; u = - grad(p)` would do the trick, plus be DAE-like?

irksome/nystrom_stepper.py
pbrubeck, line 25:
```suggestion
        return (numpy.allclose(numpy.triu(self.Abar), 0)
```

PR #130, opened 2025-03-08
Clean up UFL manipulation

pbrubeck:
This PR changes the preprocessing of `TimeDerivative` so that `Dt` gets pushed inside `Indexed`, ie. `Dt(u[i]) -> Dt(u)[i]`. This means that ultimately we only need to do replacements on `Dt(u)`, as opposed to separate replacements for each individual component `Dt(u[i])`.


rckirby (changes requested):
If this gets rid of needing the preorder `replace`, we should go through the remaining calls to this `replace` (mainly in boundary conditions) and have it point to `ufl.replace` instead of `tools.replace`?

PR #131, opened 2025-03-12
Add dDAE BCs for Nystrom

ScottMacLachlan:
When using Nystrom with an explicit Tableau, the "natural" way to do DAE BCs is to impose them based on the stage approximations of du/dt, not of u.  This introduces code to do this, in both the explicit and implicit cases.

### DIFF
--- a/demos/nystrom/demo_wave_explicit.py.rst
+++ b/demos/nystrom/demo_wave_explicit.py.rst
@@ -1,0 +1,109 @@
+Solving the Wave Equation with Irksome and an explicit Nystrom method
+=====================================================================
+
+Let's start with the simple wave equation on :math:`\Omega = [0,1]
+\times [0,1]`, with boundary :math:`\Gamma`:
+
+.. math::
+
+   u_{tt} - \Delta u &= 0
+
+   u & = 0 \quad \textrm{on}\ \Gamma
+
+At each time :math:`t`, the solution
+to this equation will be some function :math:`u\in V`, for a suitable function
+space :math:`V`.
+
+We transform this into weak form by multiplying by an arbitrary test function
+:math:`v\in V` and integrating over :math:`\Omega`.  We know have the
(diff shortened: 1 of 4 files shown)


rckirby (changes requested):
Is there a difference mathematically in DAE and dDAE in the implicit case?
And since we know "DAE" fails with explicit methods, maybe we catch that up front and give a more informative error message?

ScottMacLachlan:
Yes - in the implicit case, it's whether you're setting `u(t_n+c_i dt,x) = g(t_n + c_i dt,x)` on the boundary (in DAE) or `du/dt(t_n+c_i dt,x) = dg/dt(t_n + c_i dt,x)` (in dDAE).

The Math behind the dDAE BCs comes from the underlying partitioned scheme.  If I'm solving a second-order (in time) PDE with a partitioned scheme, I strongly impose BCs on `z = u'` as the time derivatives of the strong BCs imposed on `u` itself, and DAE BCs on `z` take the form above.  When we reduce the partitioned scheme to Nystrom, the stages that we use are those from `z`, not those from `u`, so it's consistent impose the `z` BCs within Nystrom.

ScottMacLachlan:
I've added the error message if you ask for `DAE` with an explicit Nystrom Tableau.  Just to be pendantic, I'll point out there's 3 cases here and we really only cover 2 of them:
1) Implicit, where we can invert the Abar Tableau
2) Explicit, but we could invert the shifted Tableau, like we do in dDAE and our DIRK-explicit case
3) Explicit, but both Abar and the shifted Tableau are not invertible (so you should use dDAE).  This is the case for Classic Nystrom.

I could implement case 2 inside that `if tableau.is_explicit` where that warning now sits, if it seems worthwhile to have around...

rckirby (approved):
I think this is fine for now absent a proper treatment of DIRK/explicit RK methods.   As part of that push, we probably also want to try out KMV + mass-lumping in a separate demo.  That would be a great point of comparison.

PR #132, opened 2025-03-18
TimeDerivativeRuleDispatcher: handle generic UFL types

pbrubeck:
This should extend support to time-dependent `Forms` involving generic `BaseForms` (`Cofunctions`, `ExternalOperators`, etc). Initially these generic terms are treated as time independent.

### DIFF
--- a/irksome/deriv.py
+++ b/irksome/deriv.py
@@ -79,12 +79,7 @@ class TimeDerivativeRuleDispatcher(MultiFunction):
         f, = o.ufl_operands
         return map_expr_dag(self.rules, f)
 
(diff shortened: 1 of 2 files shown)


rckirby (commented):
Is the main purpose here just to enable cofunctions (which is what is in the new test), or is more going on?

pbrubeck:
Right, with the `RuleSet` we would have thrown an error before this PR

pbrubeck:
Right, with the `RuleSet` we would have thrown an error before this PR

rckirby (approved):

PR #137, opened 2025-03-20
Fixing what we put in appctx for Rana

ScottMacLachlan:
We need to be able to call `getForm` and its variants from within `IRKAuxiliaryOperatorPC`, which requires us to stash all of the options needed to do that in our `appctx`.  This tries to make sure that we do that consistently.


rckirby (approved):
LGTM

PR #139, opened 2025-03-26
Initial integration of NPRK methods

jmbigman:
Added NPRK method and tableaux classes and a Burgers' equation demo

### DIFF
--- a/demos/nprk/demo_burgers_nprk.py.rst
+++ b/demos/nprk/demo_burgers_nprk.py.rst
@@ -1,0 +1,106 @@
+Solving the Viscous Burgers' Equation with a Non-Linearly Partitioned RK Scheme
+===============================================================================
+
+Consider the viscous Burgers' equation
+
+.. math::
+
+   \partial_t u + u\, \partial_x u - \epsilon\, \partial_x^2 u = 0
+
+To demonstrate the non-linearly partitioned RK (NPRK) method, various partitionings are listed and compared in:
+
+Buvoli, Southworth "A New Class of Runge-Kutta Methods for Nonlinearly Partitioned Systems", arXiv 2024
+
+The non-linear partitioning is explained here and implemented in Irksome.
+
+A non-linear partitioning equates the time derivative with an operator taking two arguments, :math:`u_t=F(u,u)`.
+Each stage of the NPRK scheme treats one argument implicitly and the other explicitly.
+There are many choices of non-linear partitioning, but
+
+.. math::
+
+   F(u_1, u_2) = - u_2\, \partial_x u_1 + \epsilon\, \partial_x^2 u_1
+
+is selected.
+
+With homogenous Dirichlet boundary conditions, this equation gives rise to the weak form
+
+.. math::
+
+   (v, \partial_t u) + (v, u_2\, \partial_x u_1) + (\partial_x v, \epsilon\, \partial_x u_2) = 0
+
+Following the paper, this equation is solved numerically for :math:`\epsilon=1/200`, :math:`x\in[-2,2]`, :math:`t\in[0,0.6]`
+with :math:`1000` grid points and the initial condition
+
+.. math::
+
+   u(x,0)=e^{-3x^2}
+
+The timestep :math:`\Delta t=6\cdot 10^{-3}` is selected as the middle of the range tested in the referenced paper.
+
+To implement this, begin by importing firedrake::
+
+   from firedrake import *
+
+Additionally, a few basic items and the NPRK time stepper are imported from irksome::
+
+   from irksome import Dt, MeshConstant, NPRKMethod
+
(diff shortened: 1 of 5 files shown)


demos/nprk/demo_burgers_nprk.py.rst
```diff
@@ -0,0 +1,106 @@
...
+   u = Function(V)
+   u.interpolate(u0)
+
+To define the weak form, two additional :class:`firedrake.Function`'s are needed to indicate :math:`u_1` and :math:`u_2` ::
+
+   u_1, u_2 = Function(V), Function(V)
+   EPS = Constant(1/200)
+   v = TestFunction(V)
+
+   F = inner(v, Dt(u))*dx + inner(v, u_2*u_1.dx(0))*dx \
```
rckirby, line 91: I'm concerned here about the interface relative to the additive/IMEX approach.  This has the original function `u` plus two extra functions (somehow linked) that play roles in the splitting.  I don't see a better approach short of labeling (and that might be hard multiplicatively within a single integral?)

irksome/nprk.py
```diff
@@ -0,0 +1,328 @@
...
+        _ = len(x)
+    except Exception:
+        return False
+
+    return True
+
+
+def _check_aux_t_dep(F: Form, u0: Function, u1: Function, u2: Function) -> tuple[np.ndarray]:
+    """Identifies which subfunctions are auxiliary (no time derivative)
+    and which are time-dependent. Also checks that variables are exactly
```
rckirby, line 30: "time dependent" and "no time derivative" aren't mutually exclusive.

irksome/nprk.py
```diff
@@ -0,0 +1,328 @@
...
+    one of the two and that auxiliary variable do not appear in a partitioned
+    argument."""
+
+    num_fields = len(u0.function_space())
+
+    if num_fields == 1:
```
rckirby, line 36: If you check the updated manipulation that landed not too long ago, you'll see that we've moved away from this approach to iterating over `u0` itself: `u0bits = u0[i] for i in numpy.ndindex(u.ufl_shape)`.  This is a different path that doesn't have so many branches on mixed vs vector vs whatever.

irksome/nprk.py
```diff
@@ -0,0 +1,328 @@
...
+                        and all(ub_ not in vars_ for ub_ in u0b))
+            in_args = (any(ub_ in vars_ for ub_ in u1b)
+                       and any(ub_ in vars_ for ub_ in u2b))
+        else:
+            is_aux = u0b in vars_ and TimeDerivative(u0b) not in derivs
+            is_t_dep = TimeDerivative(u0b) in derivs and u0b not in vars_
+            in_args = u1b in vars_ or u2b in vars_
+
+        if is_aux and is_t_dep:
+            raise RuntimeError(f"Variable {i} is both time-dependent and auxiliary.")
```
rckirby, line 72: `time-differentiated and auxiliary`?

irksome/nprk.py
```diff
@@ -0,0 +1,328 @@
...
+                          "bc_type": "DAE",
+                          "stage_type": "nprk",
+                          "nullspace": nullspace}
+        if appctx is None:
+            appctx = appctx_irksome
+        else:
+            appctx = {**appctx, **appctx_irksome}
+
+        # One solver corresponding to each implicit argument
+        self.solvers = {}
```
rckirby, line 224: Parallel construction -- you pass two sets of parameters with names, but then you stash the solvers in a dict.  Also, why use a dict if you're just indexing it with 0, 1?  A list/tuple would be fine.

And if you had the args come in via an iterable, creating the problems/solvers would also be done with a comprehension..

demos/nprk/demo_burgers_nprk.py.rst
```diff
@@ -0,0 +1,106 @@
...
+
+The timestep :math:`\Delta t=6\cdot 10^{-3}` is selected as the middle of the range tested in the referenced paper.
+
+To implement this, begin by importing firedrake::
+
+   from firedrake import *
+
+Additionally, a few basic items and the NPRK time stepper are imported from irksome::
+
+   from irksome import Dt, MeshConstant, NPRKMethod
```
rckirby, line 47: `NPRKMethod` doesn't match our naming for other methods.  It would be more helpful to the readier to be chattier, like `NonlinearlyPartitionedRKTimeStepper`, and more parallel to the other schemes we have.

rckirby (changes requested):
Thanks for the contribution.  Sorry for the delay in reviewing.  It's a bit to wrap one's head around since I'm unfamiliar with these methods.  I'm trying to make suggestions about programming and readability/consistency with the rest of Irksome.  More feedback might come as I further digest what you're doing.

And we, of course, need tests, too.

irksome/nprk.py
pbrubeck: We don't even have to worry about mixed vs vector these days. We just preprocess the form with `expand_time_derivatives`.

PR #142, opened 2025-04-01
WIP: make stage derivative work with cofunctions

rckirby:
We need to make sure that we can force PDE with cofunctions (e.g. delta functions) and do the right UFL replacements.  We also need to add some tests to catch this for all our steppers.

### DIFF
--- a/irksome/stage_derivative.py
+++ b/irksome/stage_derivative.py
@@ -2,7 +2,7 @@ import numpy
 from firedrake import Function, TestFunction
 from firedrake import NonlinearVariationalProblem as NLVP
 from firedrake import NonlinearVariationalSolver as NLVS
-from firedrake import assemble, dx, inner, norm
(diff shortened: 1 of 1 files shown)


tests/test_delta.py
```diff
@@ -0,0 +1,152 @@
...
+@pytest.fixture
+def vom(msh):
+    return VertexOnlyMesh(msh, [(0.5, 0.5)])
+
+
+def delta(v, vom):
+    P0 = FunctionSpace(vom, "DG", 0)
+    v_vom = TestFunction(P0)
+    F_vom = inner(1, v_vom) * dx
+    return Interpolate(v, F_vom)
```
pbrubeck:
```suggestion
    Fvom = Cofunction(P0.dual()).assign(1)
    return interpolate(v, Fvom)
```

PR #138, opened 2025-03-26
StageDerivative: add EquationBC support

ABaierReinio:
This pull request enables EquationBC support, although it currently only supports the DAE formulation.

PR #143, opened 2025-04-03
Correctly incorporate BCs that are passed to EquationBC

ABaierReinio:
In PR https://github.com/firedrakeproject/Irksome/pull/138 we missed the fact that EquationBCs can own BCs of their own, and these must also be accounted for. This is important when using EquationBCs and DirichletBCs at the same time on different parts of the boundary. This pull request incorporates that.


rckirby (approved):
Right, we forgot about bc's for the bc's.  This looks like it does the right thing.

PR #144, opened 2025-04-03
Jdstephens/fix bounds constraints

jdsteph2:
Reenable/fix bounds constraints and add an optional collocation-style update for updating the solution when using Lagrange collocation + bounds + an RK method which is not stiffly accurate.


rckirby (commented):
Was just the internal business broken, or do we need to update any of the demos?

Also, do these changes help you get the behavior you're looking for?

If so, switch the PR from draft and we'll do a more thorough review to get this merged.

jdsteph2:
Just the internals--all the demos should be alright.  And yes, everything is working as expected now.  The use of the collocation update I added is kind of obscure...it might be worth a new demo?  I can add this if it would be beneficial.

UNKNOWN.egg-info/PKG-INFO
```diff
@@ -0,0 +1,11 @@
+Metadata-Version: 2.1
```
rckirby, line 1: What is going on with these files?  Did they get generated by the pip installation?  If so, they should probably be added to `.gitignore` rather than our repository.

irksome/tools.py
```diff
@@ -9,6 +9,16 @@
 from irksome.deriv import TimeDerivative
 
 
+def flatten_dats(dats):
+    flat_dat = []
+    for dat in dats:
+        if isinstance(dat, tuple):
```
rckirby, line 15:
```suggestion
        if isinstance(dat, (tuple, list)):
```

irksome/stage_value.py
```diff
@@ -236,6 +239,22 @@ def _update_general(self):
...
+
+        assert isinstance(self.butcher_tableau, CollocationButcherTableau), "Need a collocation method for collocation update"
+        assert (self.basis_type is None or self.basis_type == "Lagrange"), "Collocation update requires the Lagrange form of the collocation polynomial"
+        assert (len(set(nodes)) == self.butcher_tableau.num_stages + 1), "Need a non-confluent collocation method to use collocation update"
+
+        lag_basis = LagrangePolynomialSet(ufc_simplex(1), nodes)
+        vander = lag_basis.tabulate(numpy.array([[1.0]]), 0)[(0,)].flatten()
+        stage_vals = numpy.insert(self.stages.subfunctions, 0, [self.u0.subfunctions[nf] for nf in range(self.num_fields)])
+
+        for i, u0bit in enumerate(self.u0.subfunctions):
```
rckirby, line 255: See DG: https://github.com/firedrakeproject/Irksome/blob/a3d5c2390589d2fad3e1fce8915dc98fe890ea8f/irksome/discontinuous_galerkin_stepper.py#L205-L209
for a way to avoid having to separate out stiffly accurate and not cases

rckirby: Also, these checks all get called at run-time (each time step), and you could lift then into the constructor.

irksome/stepper.py
```diff
@@ -57,7 +57,7 @@ def TimeStepper(F, butcher_tableau, t, dt, u0, **kwargs):
         "deriv": ["stage_type", "bcs", "nullspace", "solver_parameters", "appctx",
                   "bc_type", "splitting", "adaptive_parameters"],
         "value": ["stage_type", "basis_type", "bcs", "nullspace", "solver_parameters",
-                  "update_solver_parameters", "appctx", "splitting", "bounds"],
+                  "update_solver_parameters", "appctx", "splitting", "bounds", "use_collocation_update"],
```
rckirby, line 60: `use_collocation_update` should probably default to false, add documentation for when you can safely benefit from using it.

irksome/stepper.py
```diff
@@ -117,12 +117,14 @@ def TimeStepper(F, butcher_tableau, t, dt, u0, **kwargs):
         basis_type = kwargs.get("basis_type")
         update_solver_parameters = kwargs.get("update_solver_parameters")
         nullspace = kwargs.get("nullspace")
+        bounds = kwargs.get("bounds")
+        use_collocation_update = kwargs.get("use_collocation_update")
```
rckirby, line 121:
```suggestion
        use_collocation_update = kwargs.get("use_collocation_update", False)
```

rckirby (changes requested):
This is coming together

irksome/stage_value.py
```diff
@@ -236,6 +239,22 @@ def _update_general(self):
...
+        # Use the terminal value of the collocation polynomial to update the solution. Note: collocation update is only implemented for constant-in-time boundary conditions.
+        # TODO: create an assertion to check for constant-in-time boundary conditions.
+        nodes = numpy.insert(self.butcher_tableau.c, 0, 0.0)
+
+        assert isinstance(self.butcher_tableau, CollocationButcherTableau), "Need a collocation method for collocation update"
+        assert (self.basis_type is None or self.basis_type == "Lagrange"), "Collocation update requires the Lagrange form of the collocation polynomial"
+        assert (len(set(nodes)) == self.butcher_tableau.num_stages + 1), "Need a non-confluent collocation method to use collocation update"
+
+        lag_basis = LagrangePolynomialSet(ufc_simplex(1), nodes)
+        vander = lag_basis.tabulate(numpy.array([[1.0]]), 0)[(0,)].flatten()
```
pbrubeck, line 252: I _think_ this also works
```suggestion
        vander = lag_basis.tabulate((1.0,)))[(0,)]
```

jdsteph2: Ah, indeed. Much cleaner.

irksome/stage_value.py
```diff
@@ -186,7 +188,24 @@ def __init__(self, F, butcher_tableau, t, dt, u0, bcs=None,
...
+            assert isinstance(self.butcher_tableau, CollocationButcherTableau), "Need a collocation method for collocation update"
+            assert (self.basis_type is None or self.basis_type == "Lagrange"), "Collocation update requires the Lagrange form of the collocation polynomial"
+            assert (len(set(nodes)) == self.butcher_tableau.num_stages + 1), "Need a non-confluent collocation method to use collocation update"
+
+            lag_basis = LagrangePolynomialSet(ufc_simplex(1), nodes)
+            collocation_vander = vecconst(lag_basis.tabulate((1.0,))[(0,)])
+            stage_vals = numpy.insert(self.stages.subfunctions, 0, [self.u0.subfunctions[nf] for nf in range(self.num_fields)])
+
+            self.collocation_vander = collocation_vander
+            self.stage_vals = stage_vals
```
pbrubeck, line 205: Maybe we could get rid of `self.stage_vals` and construct the tuple each time step
```suggestion
            self.collocation_vander = collocation_vander
            self.stage_vals = self.u0.subfunctions + self.stages.subfunctions
```

irksome/stage_value.py
jdsteph2: I moved the checks and algebra into the constructor.

If I'm not mistaken, a general method can be stiffly accurate and confluent. The basis tabulation breaks down in this case. Would we then want to separate the confluent update into its own case?

UNKNOWN.egg-info/PKG-INFO
jdsteph2: Not sure exactly where these came from.

irksome/stage_value.py
jdsteph2: Agreed. This cleans things up nicely.

irksome/stage_derivative.py
```diff
@@ -115,7 +115,8 @@ def bc2stagebc(bc, i):
                 F_bc_orig = expand_time_derivatives(bc.f, t=t, timedep_coeffs=(u0,))
                 F_bc_new = replace(F_bc_orig, repl[i])
                 Vbigi = stage2spaces4bc(bc, V, Vbig, i)
-                return EquationBC(F_bc_new == 0, stages, bc.sub_domain, V=Vbigi)
+                return EquationBC(F_bc_new == 0, stages, bc.sub_domain, V=Vbigi,
+                                  bcs=[bc2stagebc(innerbc, i) for innerbc in extract_bcs(bc.bcs)])
```
pbrubeck, line 119: Could you `git checkout master; git pull; git checkout jdstephens/fix_bounds_constraints; git merge master`?

irksome/stepper.py
jdsteph2: I added a note about the kwarg under the base time stepper. I can move it elsewhere if there is a better spot for it.

irksome/stage_derivative.py
jdsteph2: Done!

irksome/base_time_stepper.py
```diff
@@ -87,6 +85,12 @@ class StageCoupledTimeStepper(BaseTimeStepper):
     :arg butcher_tableau: A :class:`ButcherTableau` instance giving
             the Runge-Kutta method to be used for time marching.
     :arg bounds: An optional kwarg used in certain bounds-constrained methods.
+    :arg use_collocation_update: An optional kwarg indicating whether to use
+        the terminal value of the collocation polynomial as the solution
+        update. This is needed to bypass the mass matrix inversion when
+        enforcing bounds constraints with an RK method that is not stiffly
+        accurate. Currenty, only constant-in-time boundary conditions are
```
pbrubeck, line 92:
```suggestion
        accurate. Currently, only constant-in-time boundary conditions are
```

irksome/tools.py
```diff
@@ -5,10 +5,23 @@
...
 
 from irksome.deriv import TimeDerivative
 
 
+def flatten_dats(dats):
+    flat_dat = []
+    for dat in dats:
+        if isinstance(dat, MixedDat):
+            flat_dat.extend(dat.split)
+        elif isinstance(dat, (tuple, list)):
```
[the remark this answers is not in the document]
pbrubeck: It might be cleaner to `append` if we have a `Dat`, or `extend` otherwise.

irksome/stage_value.py
rckirby: We're already not handling the confluent case in code (there would be another mass matrix inversion floating around somewhere?).  For now, you could just check on input that you shouldn't have an honest-to-goodness `CollocationButcherTableau` with `c[0]` as zero.

irksome/stepper.py
rckirby: Maybe it belongs on the stage value stepper as well since that's the only place we can use it?

rckirby (commented):
Please go ahead and deal with Pablo's suggestions when you get the chance.  I think we're getting close.

irksome/tools.py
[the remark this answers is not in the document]
jdsteph2: Ah, much cleaner indeed!

irksome/stage_value.py
jdsteph2: Ah, I see.  I think this is in a better spot now. What I have currently,
`assert (len(set(nodes)) == self.butcher_tableau.num_stages + 1)`,
should check for the case when c[0] = 0, and c[i] = c[j] elsewhere.

irksome/base_time_stepper.py
```diff
@@ -1,10 +1,8 @@
-import numpy
-
 from abc import abstractmethod
 from firedrake import Function, NonlinearVariationalProblem, NonlinearVariationalSolver
 from firedrake.dmhooks import pop_parent, push_parent
 from firedrake.petsc import PETSc
-from .tools import AI, get_stage_space, getNullspace
+from .tools import AI, get_stage_space, getNullspace, flatten_dats
 from pyop2.types import MixedDat
```
pbrubeck, line 6: Almost there. Could `flatten_dats` return a `MixedDat`? In this way we only import `MixedDat` in tools.py.

jdsteph2: Done!

irksome/stage_value.py
rckirby: What's the particular error?  It's probably something subtle at the interface of UFL + numpy...

irksome/stepper.py
rckirby: Ok, maybe just do it on the `TimeStepper` function then and say it only make sense for stage value?

irksome/stage_value.py
pbrubeck: @jdsteph2 could you add a test?

jdsteph2: > @jdsteph2 could you add a test?

Certainly. I'll clean something up and add an additional test.

irksome/stage_value.py
```diff
@@ -236,6 +253,11 @@ def _update_general(self):
         self.update_solver.solve()
         self.u0.assign(self.unew)
 
+    def _update_collocation(self):
+        stage_vals = self.u0.subfunctions + self.stages.subfunctions
+        for i, u0bit in enumerate(self.u0.subfunctions):
+            u0bit.assign(numpy.dot(self.collocation_vander[:], stage_vals[i::self.num_fields]))
```
pbrubeck, line 259: Maybe this is the fix you need
```suggestion
            u0bit.assign(numpy.dot(stage_vals[i::self.num_fields]. self.collocation_vander))
```

pbrubeck: This should be fixed now

irksome/stage_value.py
pbrubeck: presumably the DG-in-time update is also broken

pbrubeck: I also fixed DG-in-time

pbrubeck (approved):
Thanks for all the work @jdsteph2

PR #145, opened 2025-04-29
Propagate kwargs to NLVP/NLVS

pbrubeck:


rckirby (approved):
Thanks, I appreciate that you picked up more kwargs that I'd never gotten around to setting, too!

PR #146, opened 2025-04-30
steppers go into appctx

rckirby:
This lets us streamline a lot of business in preconditioners that re-discretize in time.

As a further simplification, we could remove everything that's in the appctx that is also a member of the class.  Not sure that wouldn't upset anybody?

### DIFF
--- a/irksome/base_time_stepper.py
+++ b/irksome/base_time_stepper.py
@@ -26,7 +26,8 @@ class BaseTimeStepper:
             "dt": dt,
             "u0": u0,
             "bcs": bcs,
-            "nullspace": nullspace}
+            "nullspace": nullspace,
+            "stepper": self}
 
         if appctx is None:
             self.appctx = appctx_base
@@ -155,7 +156,7 @@ class StageCoupledTimeStepper(BaseTimeStepper):
     # allow butcher tableau as input for preconditioners to create
     # an alternate operator
     @abstractmethod
-    def get_form_and_bcs(self, stages, butcher_tableau=None):
+    def get_form_and_bcs(self, stages, tableau=None, F=None):
         pass
 
     def solver_stats(self):
--- a/irksome/base_time_stepper.py
+++ b/irksome/base_time_stepper.py
@@ -20,14 +20,7 @@ class BaseTimeStepper:
         self.nullspace = nullspace
         self.V = u0.function_space()
 
-        appctx_base = {
-            "F": F,
-            "t": t,
-            "dt": dt,
-            "u0": u0,
-            "bcs": bcs,
-            "nullspace": nullspace,
-            "stepper": self}
+        appctx_base = {"stepper": self}
 
         if appctx is None:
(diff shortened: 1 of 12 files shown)


irksome/base_time_stepper.py
```diff
@@ -26,7 +26,8 @@ def __init__(self, F, t, dt, u0,
             "dt": dt,
             "u0": u0,
             "bcs": bcs,
-            "nullspace": nullspace}
+            "nullspace": nullspace,
+            "stepper": self}
```
ScottMacLachlan, line 30: This feels redundant - `stepper` already includes `F`, `t`, `dt`, `u0`, `nullspace`, and `bcs` as data, so we now stash those pieces twice.  Is there a good design reason to include things both directly and indirectly?

irksome/imex.py
```diff
@@ -216,6 +220,7 @@ def __init__(self, F, Fexp, butcher_tableau,
                           "u0": u0,
                           "bcs": bcs,
                           "stage_type": "value",
+                          "stepper": self,
                           "splitting": splitting,
                           "nullspace": nullspace}
```
ScottMacLachlan, line 225: Same comment as above, that we're double-stashing information.

irksome/pc.py
```diff
@@ -54,16 +49,11 @@ def getAtilde(self, A):
...
-        bcs = appctx["bcs"]
-        stage_type = appctx.get("stage_type", None)
-        bc_type = appctx.get("bc_type", None)
-        splitting = appctx.get("splitting", None)
-        vandermonde = appctx.get("vandermonde", None)
+        stepper = appctx["stepper"]
+        butcher = stepper.butcher_tableau
+        F = stepper.F
+        u0 = stepper.u0
+        bcs = stepper.orig_bcs
```
ScottMacLachlan, line 56: As above, now we're pulling things indirectly out of `stepper`, not directly from the `appctx`, so we should reconsider what we're putting in there.

ScottMacLachlan (commented):
Overall, this looks good.  I just have a small (recurring) concern about redundancy in what we stash in the appctx

irksome/base_time_stepper.py
rckirby: So you discovered this trick, too.  The question is whether removing them fro the appctx would harm any users who are not politely obeying the public interface.

irksome/pc.py
rckirby: Agreed.  Do we need actually need anything besides the stepper in the app context?

ScottMacLachlan: We currently stash `splitting` and `bc_type` in `StageCoupledTimeStepper`, and some similar bits elsewhere.  But it looks to me like all of these are also stashed in `stepper` directly.  So, I'd say we go the simple route and just stash `stepper`...

ScottMacLachlan (approved):
LGTM. Thanks!

PR #148, opened 2025-05-14
WIP: SDC python type SNES

JHopeCollins:


demos/sdc_burgers.py
```diff
@@ -0,0 +1,60 @@
...
+        "rtol": 1e-5,
+    },
+    "snes_type": "python",
+    "snes_python_type": "irksome.SDCLDSNES",
+    "sdc": newton_params,
+    "sdc": {
+        "snes_type": "python",
+        "snes_python_type": "firedrake.FieldsplitSNES",
+        "snes_fieldsplit_type": "multiplicative",
+        "fieldsplit_0": newton_params,
```
rckirby, line 50: Doesn't `"fieldsqplit": newton_params` just broadcast the options into all splits?  This is true for field splits inside of PC.

PR #157, opened 2025-07-16
Pipe bounds into Galerkin-in-time methods

jdsteph2:
Initial work on piping bounds into Galerkin-in-time methods

### DIFF
--- a/irksome/base_time_stepper.py
+++ b/irksome/base_time_stepper.py
@@ -167,7 +167,7 @@ class StageCoupledTimeStepper(BaseTimeStepper):
         if upper is None:
             sub = Function(Vbig).assign(PETSc.INFINITY)
 
-        if bounds_type == "stage":
+        if (bounds_type == "stage") or (bounds_type == "galerkin"):
             if lower is not None:
                 dats = [lower.dat] * (self.num_stages)
                 slb = Function(Vbig, val=flatten_dats(dats))
--- a/irksome/discontinuous_galerkin_stepper.py
+++ b/irksome/discontinuous_galerkin_stepper.py
@@ -177,9 +177,10 @@ class DiscontinuousGalerkinTimeStepper(StageCoupledTimeStepper):
             instance to be passed to a
             `firedrake.MixedVectorSpaceBasis` over the larger space
             associated with the Runge-Kutta method
+    :arg bounds: An optional kwarg used in certain bounds-constrained methods.
     """
     def __init__(self, F, order, t, dt, u0, bcs=None, basis_type=None,
-                 quadrature=None, **kwargs):
+                 quadrature=None, bounds=None, **kwargs):
         assert order >= 0
         self.order = order
         self.basis_type = basis_type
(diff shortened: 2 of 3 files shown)


irksome/discontinuous_galerkin_stepper.py
```diff
@@ -198,7 +199,10 @@ def __init__(self, F, order, t, dt, u0, bcs=None, basis_type=None,
 
         self.update_b = vecconst(self.el.tabulate(0, (1.0,))[(0,)])
 
-        super().__init__(F, t, dt, u0, num_stages, bcs=bcs, **kwargs)
+        if bounds is not None:
+            assert (basis_type is None) or (basis_type == "Lagrange") or (basis_type == "Bernstein")
```
pbrubeck, line 203:
```suggestion
            assert (basis_type is None) or (basis_type != "integral")
```

pbrubeck: We should be raising a `ValueError` otherwise

irksome/base_time_stepper.py
```diff
@@ -167,7 +167,7 @@ def get_stage_bounds(self, bounds=None):
         if upper is None:
             sub = Function(Vbig).assign(PETSc.INFINITY)
 
-        if bounds_type == "stage":
+        if (bounds_type == "stage") or (bounds_type == "galerkin"):
```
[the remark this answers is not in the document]
ScottMacLachlan: `collocation`?

irksome/discontinuous_galerkin_stepper.py
jdsteph2: While much cleaner, I figured that we should err on the side of caution and explicitly allow basis types when working with bounds. Then, one has to think about whether or not bounds make sense before you can impose them with any new basis type. If we don't anticipate non-integral non-pointwise basis types, then checking `basis_type != "integral"` should be good.

rckirby: I'm fine with that option.  If we hit other cases where this breaks, we can fix it.  I'll just commit Pablo's suggestion.

tests/test_galerkinBounds.py
```diff
@@ -0,0 +1,328 @@
...
+
+        if min_val < -1.0:
+            bounds_violations.append(min_val)
+        if max_val > 1.0:
+            bounds_violations.append(max_val)
+
+    return bounds_violations
+
+
+@pytest.mark.parametrize('order', (1, 2, 3))
```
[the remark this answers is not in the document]
rckirby: You're not testing GLL -- you should have `basis_type = "gll"` instead of just `"Lagrange"` to try that out.  My conjecture is that this *with* GLL quadrature should work, Bernstein with any (accurate enough) quadrature should work, and that anything else might or might not.  I would be mostly happy with this PR if that were the case.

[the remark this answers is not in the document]
jdsteph2: Sorry, I forgot to push the updated tests...new failing tests should be incoming.

PR #158, opened 2025-07-25
Fix Tableau typo

ScottMacLachlan:
There's a typo in our Classic Nystrom Tableau, in comparison to that given in Hairer-Norsett-Wanner (see attached screenshot).  Let's maybe fix that...

screenshot_2025-07-25_at_8 13 28___pm_720


rckirby (approved):
Oops, and thanks!

PR #159, opened 2025-07-27
Useful error message

ScottMacLachlan:
When an invalid kwarg is passed to `TimeStepper`, we catch it and error out, but don't provide a useful message about which kwarg is invalid.  This fixes that oversight.


rckirby (approved):
Sure, that's helpful.

PR #172, opened 2025-10-29
Move galerkin steppers behind TimeStepper interface

rckirby:
I've added `DGDescriptor` and `CPGDescriptor` that encapsulate the order and basis/quadrature kwargs we were passing to `GalerkinTimeStepper` and `DiscontinuousGalerkinTimeStepper` so that we can pass those to `TimeStepper` in the slot RK methods use a butcher tableau for -- now, it's easier to switch between RK and Galerkin.

### DIFF
--- a/irksome/stepper.py
+++ b/irksome/stepper.py
@@ -21,7 +21,9 @@ valid_kwargs_per_stage_type = {
     "explicit": ["stage_type", "bcs", "solver_parameters", "appctx"],
     "imex": ["Fexp", "stage_type", "it_solver_parameters", "prop_solver_parameters",
              "splitting", "num_its_initial", "num_its_per_step"],
-    "dirkimex": ["Fexp", "stage_type", "mass_parameters"]}
+    "dirkimex": ["Fexp", "stage_type", "mass_parameters"],
+    "dg": ["bcs"],
+    "cpg": ["aux_indices", "bcs"]}
 
 valid_adapt_parameters = ["tol", "dtmin", "dtmax", "KI", "KP",
                           "max_reject", "onscale_factor",
@@ -100,13 +102,12 @@ def TimeStepper(F, method, t, dt, u0, **kwargs):
             auxiliary, that is, discretized in the lower-order DG test space.
     """
     # first pluck out the cases for Galerkin in time...
-    valid_dg_kwargs = ["bcs", "solver_parameters", "appctx", "nullspace"]
-    valid_cpg_kwargs = valid_dg_kwargs + ["aux_indices"]
+
     if isinstance(method, DiscontinuousGalerkinScheme):
-        assert set(kwargs.keys()).issubset(valid_dg_kwargs)
+        assert set(kwargs.keys()).issubset(list(valid_base_kwargs) + valid_kwargs_per_stage_type["dg"])
         return DiscontinuousGalerkinTimeStepper(F, method, t, dt, u0, **kwargs)
     elif isinstance(method, ContinuousPetrovGalerkinScheme):
-        assert set(kwargs.keys()).issubset(valid_cpg_kwargs)
+        assert set(kwargs.keys()).issubset(list(valid_base_kwargs) + valid_kwargs_per_stage_type["cpg"])
         return ContinuousPetrovGalerkinTimeStepper(F, method, t, dt, u0, **kwargs)
 
     stage_type = kwargs.pop("stage_type", "deriv")


irksome/__init__.py
```diff
@@ -29,6 +29,7 @@
 from .stepper import TimeStepper            # noqa: F401
 from .tools import MeshConstant             # noqa: F401
 from .wso_dirk_tableaux import WSODIRK      # noqa: F401
+from .descriptor import CPGDescriptor, DGDescriptor  # noqa: F401
```
pbrubeck: This file has not been added to the repo!

rckirby:
> Looking ahead, it'd be good if these Descriptor classes are generic enough to support Basix elements.

I would be glad to support basix -- If the Descriptor is completely non-numerical (we'd need to use some kind of string/label for quadrature instead of an actual rule), we could have a widget that supplies the basis/quadrature rule from either FIAT or basix as desired.  That argues for generalizing how we do quadrature rules as well.

irksome/descriptor.py
```diff
@@ -0,0 +1,30 @@
...
+    :arg order: An integer indicating the order of the method
+    :kwarg basis_type: A string indicating the finite element family (either
+           `'Lagrange'` or `'Bernstein'`) or the Lagrange variant for the
+           test/trial spaces. Defaults to equispaced Lagrange elements.
+    :kwarg quadrature: A :class:`FIAT.QuadratureRule` indicating the quadrature
+            to be used in time, defaulting to GL with order points
+    """
+    def __init__(self, order, basis_type, quadrature):
+        self.order = order
+        self.basis_type = basis_type
```
rckirby: For greater generality, we could also allow separate types for test and trial bases?

pbrubeck:
What about more intuitive names: `GalerkinScheme` / `DiscontinuousGalerkinScheme`, the term "descriptor" is too generic/vague

pbrubeck:
> But in Firedrake we stash quadrature on the `dx`, and we don't have an analogous `dt` thing here to support the semidiscrete interface and interoperability of such forms with RK + Galerkin.  So the quadrature rule has to live somewhere.  We cut against that somewhat with `QuadratureLabel`, however...

In Firedrake, the default quadrature rule doesn't live anywhere. It is only materialized during compile time when we do the degree estimation. 

In Irksome we can label each term. Unlabeled terms should receive a similar degree estimation. This is feasible, we need to implement a DAG visitor that applies the same degree estimation logic.

irksome/stepper.py
```diff
@@ -92,6 +95,12 @@ def TimeStepper(F, butcher_tableau, t, dt, u0, **kwargs):
         accurate. Currently, only constant-in-time boundary conditions are
         supported.
     """
+    # first pluck out the cases for Galerkin in time...
+    if isinstance(butcher_tableau, ContinuousPetrovGalerkinScheme):
```
pbrubeck: The names here are inconsistent

pbrubeck: Should `GalerkinTimeStepper` become `ContinuousPetrovGalerkinTimeStepper`?

tests/test_delta.py
```diff
@@ -94,7 +94,7 @@ def heat_delta(msh, vom, bt, stage_type):
     stepper.advance()
 
 
-def heat_delta_galerkin(msh, vom, stepper, order):
+def heat_delta_galerkin(msh, vom, descriptor):
```
pbrubeck:
```suggestion
def heat_delta_galerkin(msh, vom, scheme):
```

irksome/scheme.py
```diff
@@ -0,0 +1,30 @@
...
+    :kwarg basis_type: A string indicating the finite element family (either
+           `'Lagrange'` or `'Bernstein'`) or the Lagrange variant for the
+           test/trial spaces. Defaults to equispaced Lagrange elements.
+    :kwarg quadrature: A :class:`FIAT.QuadratureRule` indicating the quadrature
+            to be used in time, defaulting to GL with order points
+    """
+    def __init__(self, order, basis_type, quadrature):
+        self.order = order
+        self.basis_type = basis_type
+        self.quadrature = quadrature
```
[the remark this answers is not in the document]
pbrubeck: Both the basis and the quadrature are (so far) described by `(str, int)`. Maybe we should have something like
```
def __init__(basis_degree, basis_type=None, quadrature_degree=None, quadrature_scheme="deafult")
```
It would be a pain for users to manually construct FIAT quadrature rules

demos/preconditioning/demo_heat_mg_dg.py.rst
```diff
@@ -97,8 +97,8 @@ monolithic multigrid with pointwise block Jacobi preconditioning::
   
 These solver parameters work just fine in the :class:`.DiscontinuousGalerkinTimeStepper`::
```
pbrubeck:
```suggestion
These solver parameters work just fine using a :class:`.DiscontinuousGalerkinScheme`::
```

irksome/discontinuous_galerkin_stepper.py
```diff
@@ -178,19 +173,21 @@ class DiscontinuousGalerkinTimeStepper(StageCoupledTimeStepper):
...
+        self.basis_type = basis_type = scheme.basis_type
 
         V = u0.function_space()
         self.num_fields = len(V)
 
         self.el = getElement(basis_type, order)
 
-        if quadrature is None:
-            quadrature = make_quadrature(ufc_line, order+1)
+        quad_degree = scheme.quadrature_degree or 2 * (order+1)
```
pbrubeck:
```suggestion
        quad_degree = scheme.quadrature_degree
        if quad_degree is None:
            quad_degree = 2 * order
```

irksome/galerkin_stepper.py
```diff
@@ -196,20 +189,19 @@ class GalerkinTimeStepper(StageCoupledTimeStepper):
...
         V = u0.function_space()
         self.num_fields = len(V)
 
         self.trial_el, self.test_el = getElements(basis_type, order)
 
-        if quadrature is None:
-            quad_degree = self.trial_el.degree() + self.test_el.degree()
-            quadrature = create_quadrature(ufc_line, quad_degree)
+        quad_degree = scheme.quadrature_degree or self.trial_el.degree() + self.test_el.degree()
+        quadrature = create_quadrature(ufc_line, quad_degree, scheme.quadrature_scheme)
```
```suggestion
        if isinstance(scheme.quadrature_scheme, str):
            quad_degree = scheme.quadrature_degree
            if quad_degree is None:
                quad_degree = self.trial_el.degree() + self.test_el.degree()
            quadrature = create_quadrature(ufc_line, quad_degree, scheme.quadrature_scheme)
        else:
            quadrature = scheme.quadrature_scheme
```

tests/test_disc_galerkin.py
```diff
@@ -149,8 +152,10 @@ def test_1d_heat_homogeneous_dirichletbc(order):
     ufc_line = FIAT.ufc_simplex(1)
     quadrature = FIAT.quadrature.RadauQuadratureLineRule(ufc_line, order+1)
 
-    stepper = DiscontinuousGalerkinTimeStepper(
-        F, order, t, dt, u, bcs=bcs, quadrature=quadrature,
+    scheme = DiscontinuousGalerkinScheme(order, None, quadrature)
```
pbrubeck:
```suggestion
    scheme = DiscontinuousGalerkinScheme(order, quadrature_scheme=quadrature)
```

irksome/scheme.py
```diff
@@ -0,0 +1,51 @@
...
+                 quadrature_degree=None, quadrature_scheme="default"):
+        assert order >= 1, "CPG must have order >= 1"
+        super().__init__(order, basis_type,
+                         quadrature_degree, quadrature_scheme)
+
+
+ufc_line = ufc_simplex(1)
+def scheme2quadrature(quadrature_degree, quadrature_scheme):
+    if "quadrature_scheme" == "default":
+        return create_quadrature(ufc_simplex, quadrature_degree)
```
pbrubeck: Should we just pass the scheme here (unless it is "radau")

rckirby: Yes, this is a big mistake (comparing string rather than value).  We should intercept radau (& anything else that we want to support) and then default by passing `quadrature_scheme` into `create_quadrature`.

tests/test_galerkin.py
```diff
@@ -243,8 +247,10 @@ def galerkin_wave(n, deg, alpha, order):
 
     E = 0.5 * (inner(u, u)*dx + inner(p, p)*dx)
 
-    stepper = GalerkinTimeStepper(F, order, t, dt, up,
-                                  solver_parameters=params)
+    scheme = ContinuousPetrovGalerkinScheme(order)
+
```
pbrubeck:
```suggestion
```

tests/test_galerkin.py
```diff
@@ -71,8 +69,8 @@ def test_1d_heat_dirichletbc(order, basis_type):
 
 
 @pytest.mark.parametrize("order", [1, 2, 3])
-@pytest.mark.parametrize("quad_degree", [None, 6])
-def test_1d_heat_neumannbc(order, quad_degree):
+@pytest.mark.parametrize("quad_degree_raise", [0, 2])
+def test_1d_heat_neumannbc(order, quad_degree_raise):
```
pbrubeck:
```suggestion
def test_1d_heat_neumannbc(order, quad_degree):
```

tests/test_galerkin.py
```diff
@@ -99,10 +97,8 @@ def test_1d_heat_neumannbc(order, quad_degree):
 
     luparams = {"mat_type": "aij", "ksp_type": "preonly", "pc_type": "lu"}
 
-    ufc_line = ufc_simplex(1)
-
-    scheme = ContinuousPetrovGalerkinScheme(
-        order, quadrature_degree=quad_degree)
+    quad_degree = 2*order-1 + quad_degree_raise
```
pbrubeck:
```suggestion
```

irksome/stepper.py
```diff
@@ -92,6 +95,12 @@ def TimeStepper(F, butcher_tableau, t, dt, u0, **kwargs):
         accurate. Currently, only constant-in-time boundary conditions are
         supported.
     """
+    # first pluck out the cases for Galerkin in time...
```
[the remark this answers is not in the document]
rckirby: `method` instead since we've overloaded `scheme` somewhat?

irksome/stepper.py
```diff
@@ -94,12 +95,19 @@ def TimeStepper(F, butcher_tableau, t, dt, u0, **kwargs):
         enforcing bounds constraints with an RK method that is not stiffly
         accurate. Currently, only constant-in-time boundary conditions are
         supported.
+    :kwarg aux_indices: Only valid for continuous Petrov Galerkin time scheme.  It
+            specifies that some of the variables in `u0` are to be treated as
+            auxiliary, that is, discretized in the lower-order DG space.
```
pbrubeck:
```suggestion
            auxiliary, that is, discretized in the lower-order DG test space.
```

irksome/stepper.py
```diff
@@ -94,12 +95,19 @@ def TimeStepper(F, butcher_tableau, t, dt, u0, **kwargs):
...
     """
     # first pluck out the cases for Galerkin in time...
-    if isinstance(butcher_tableau, ContinuousPetrovGalerkinScheme):
-        return ContinuousPetrovGalerkinTimeStepper(F, butcher_tableau, t, dt, u0, **kwargs)
-    elif isinstance(butcher_tableau, DiscontinuousGalerkinScheme):
-        return DiscontinuousGalerkinTimeStepper(F, butcher_tableau, t, dt, u0, **kwargs)
+    valid_dg_kwargs = ["bcs", "solver_parameters", "appctx", "nullspace"]
+    valid_cpg_kwargs = valid_dg_kwargs + ["aux_indices"]
```
pbrubeck: These live at the top of the file

tests/test_galerkin.py
```diff
@@ -330,9 +324,11 @@ def kepler_aux_variable(V, order, t, dt, u0, solver_parameters):
 
     # Auxiliary variable subspaces
     aux_indices = list(range(1, len(Z)))
-    stepper = GalerkinTimeStepper(F, order, t, dt, z,
-                                  solver_parameters=solver_parameters,
-                                  aux_indices=aux_indices)
+
+    scheme = ContinuousPetrovGalerkinScheme(order)
```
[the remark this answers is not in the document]
ScottMacLachlan: So, my question is whether the only auxiliary scheme that will ever make sense is to discretize these fields in the test space of the CPG scheme, or if we we will want to be able to specify their discretization in another space (e.g., the corresponding Bernstein test space if we want to impose pointwise constraints on these fields, even if we don't want them on the primal fields).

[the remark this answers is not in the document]
rckirby: There is a separate issue (separate PR) of allowing different "families" for test and trial bases, but right now, one has that variables are by default in the CG space of degree k or, if tagged `auxiliary` the DG space of degree k-1.  Those are the only two choices of which I'm aware.

irksome/scheme.py
```diff
@@ -0,0 +1,30 @@
...
+    :arg order: An integer indicating the order of the method
+    :kwarg basis_type: A string indicating the finite element family (either
+           `'Lagrange'` or `'Bernstein'`) or the Lagrange variant for the
+           test/trial spaces. Defaults to equispaced Lagrange elements.
+    :kwarg quadrature: A :class:`FIAT.QuadratureRule` indicating the quadrature
+            to be used in time, defaulting to GL with order points
+    """
+    def __init__(self, order, basis_type, quadrature):
+        self.order = order
+        self.basis_type = basis_type
```
ScottMacLachlan: This feels like a thing we should allow, even if it defaults to being the same

ScottMacLachlan (approved):
This looks good to me.  Thanks!

PR #176, opened 2025-11-04
Adjoint Taylor test for the Hessian

JHopeCollins:
Adds a Taylor test to check that the Hessian calculation is correct.

Depends on ~~https://github.com/firedrakeproject/firedrake/pull/4705~~ next Firedrake release.

Also removes warnings raised by firedrake adjoint when assigning an integer to a Function.


rckirby (commented):
Can we point the CI somehow to the next firedrake release to check out the failing tests?

JHopeCollins:
We now build a Docker container on each push to the Firedrake `release` and `main` branches, so we can test against those.

The tests do pass offline for me with the fix in the linked Firedrake PR.

PR #177, opened 2025-11-05
Space-time measures (e.g. dx_time/ds_time) to simplify Galerkin schemes

BorisAndrews:
This was intended to simplify the implementation of Galerkin-in-time, but I believe it clashes with @pbrubeck's [push yesterday](https://github.com/firedrakeproject/Irksome/pull/172) (since this renders TimeQuadratureLabel objects unnecessary in the definition of the form).

Naturally, Pablo's PR is nicer---it aligns more with the rest of the schemes in Irksome, and is much more user friendly---but I'd highlight that one doesn't necessarily need/want the same time quadrature on all integrands in practice. (For example, in conservative methods, you only need exact quadrature when defining your AVs. The rest can use whatever quadrature you want, which can very much speed up assembly.) I'm not sure this is possible any more with Galerkin-in-time hidden behind the TimeStepper interface?

I don't know if this lack of support is super important. I thought I'd leave the PR in regardless, in case the material is useful, but feel free to close if appropriate. :)

### DIFF
--- a/irksome/tools.py
+++ b/irksome/tools.py
@@ -3,11 +3,13 @@ from functools import reduce
 import numpy
 from firedrake import Function, FunctionSpace, VectorSpaceBasis, MixedVectorSpaceBasis, Constant, split
 from ufl.algorithms.analysis import extract_type
+from ufl.measure import Measure, MeasureSum, integral_type_to_measure_name, measure_name_to_integral_type
 from ufl import as_tensor, zero
 from ufl import replace as ufl_replace
 from pyop2.types import MixedDat
 
 from irksome.deriv import TimeDerivative
+from irksome.labeling import TimeQuadratureLabel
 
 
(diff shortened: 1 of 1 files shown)


rckirby (changes requested):
Tests/demos highlighting the usage would be quite helpful?

pbrubeck:
There is potential for confusion if we start supporting two interfaces or even if `TimeMeasure` becomes a wrapper for `TimeQuadratureLabel` (as the PR proposes). 

Should `GalerkinTimeSteppers` expect symbolic time integrals, or should they take care of that themselves? 

What does `integrand1 * dx + integrand2 * dx_time` mean? If `dx_time` is just wrapping a `TimeQuadratureLabel`, then the `TimeStepper` will integrate everything in time. Preventing these expressions to be formed might seem difficult.

PR #178, opened 2025-11-07
GalerkinCollocationScheme

pbrubeck:
Implements stage-value and stage-derivative collocation schemes for `ContinuousPetrovGalerkinTimeStepper`

Enables collocation Gauss-Legendre trial/test spaces on the default quadrature (plus the left endpoint)
```
GalerkinCollocationScheme(order, stage_type="value")

GalerkinCollocationScheme(order, stage_type="deriv")
```
Also enables collocation Gauss-Radau trial/test spaces that default to Gauss-Radau quadrature schemes
```
GalerkinCollocationScheme(order, quadrature_scheme="radau")
```

### DIFF
--- a/irksome/labeling.py
+++ b/irksome/labeling.py
@@ -36,12 +36,12 @@ class TimeQuadratureRule:
 
 def split_quadrature(F, Qdefault=None):
     """Splits a :class:`LabelledForm` into the terms to be integrated in time by the
-    different :class:`TimeQuadratureRule`s.
+    different :class:`TimeQuadratureRule` objects used as labels.
 
     :arg F: a :class:`LabelledForm`.
     :kwarg Qdefault: the :class:`TimeQuadratureRule` to be applied on unlabelled terms.
 
-    :returns: a `dict` mapping unique :class:`TimeQuadratureRule`s to :class:`Form`s.
+    :returns: a `dict` mapping unique :class:`TimeQuadratureRule` objects to the :class:`Form` to be integrated in time.
     """
     if not isinstance(F, LabelledForm):
         return {Qdefault: F}


tests/test_galerkin.py
```diff
@@ -7,7 +7,7 @@
 
 
 @pytest.mark.parametrize("order", [1, 2, 3])
-@pytest.mark.parametrize("basis_type", ["Lagrange", "Bernstein", "integral"])
+@pytest.mark.parametrize("basis_type", ["Lagrange", "Bernstein", "integral", ("enriched", "spectral")])
```
rckirby: What's the right incantation to have this test an enriched version that uses `FIAT.GaussRadau`?

rckirby: This is the cPG quadrature variant to produce something that is exactly the RadauIIA RK method.

irksome/discontinuous_galerkin_stepper.py
```diff
@@ -13,15 +13,19 @@
 
 
 def getElement(basis_type, order):
+    if basis_type is not None:
+        basis_type = basis_type.lower()
     if order == 0:
         return DiscontinuousLagrange(ufc_line, order)
-    elif basis_type == "Bernstein":
+    elif basis_type == "bernstein":
```
rckirby: Do we want to `lower()` to support either variant? same for other strings.

irksome/galerkin_stepper.py
```diff
@@ -1,48 +1,45 @@
...
 def getElements(basis_type, order):
     if isinstance(basis_type, (tuple, list)):
         trial_type, test_type = basis_type
     else:
         trial_type = basis_type
         test_type = basis_type
-    L_trial = getTrialElement(trial_type, order)
+
     L_test = getTestElement(test_type, order-1)
+    if trial_type == "enriched":
```
rckirby: Not sure "enriched" is how I would name this mathematically absent the technical naming within FIAT, but I really don't have a better idea at the present time.

tests/test_galerkin.py
```diff
@@ -50,7 +54,8 @@ def test_1d_heat_dirichletbc(order, basis_type):
 
     luparams = {"mat_type": "aij", "ksp_type": "preonly", "pc_type": "lu", "pc_factor_mat_solver_type": "mumps"}
 
-    scheme = ContinuousPetrovGalerkinScheme(order, basis_type)
+    quad_scheme = "radau" if isinstance(basis_type, tuple) and basis_type[1] == "radau" else "default"
```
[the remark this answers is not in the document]
rckirby: Depends -- do we want to default to the equivalent collocation, or default to something accurate enough for linear terms?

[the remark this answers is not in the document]
rckirby: I suppose that all bases are about the same if you're integrating to high order accuracy, so that if you care about specifying which basis you may be aiming to get special properties, so defaulting to the most likely choice of a quadrature rule (the collocating one) might make sense?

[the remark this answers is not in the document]
rckirby: For DG, I haven't worked out the connection between basis choices and existing algorithms (other than DG(0) with right-hand quadrature is backward Euler).  Maybe we want to leave off DG from this PR until we understand what it is we want to expose, then do the right thing later?  I do think that Radau should be on the list of things that give us something useful for DG.

rckirby (approved):

irksome/scheme.py
```diff
@@ -49,16 +49,46 @@ class ContinuousPetrovGalerkinScheme(GalerkinScheme):
...
+            raise ValueError(f"Unsupported stage type {stage_type}.")
+
+        trial_type = stage_type
+        basis_type = (trial_type, test_type)
+        super().__init__(order, basis_type=basis_type,
+                         quadrature_degree=quadrature_degree,
+                         quadrature_scheme=quadrature_scheme)
+
+
+def create_time_quadrature(degree, scheme=None):
```
rckirby: It might be clearer to use `num_points` rather than degree and then instantiate the Radau (or Gauss) with that number of points, independent of the degree we're after.  That, rather than order of accuracy, is the main thing in the structure here.

tests/test_disc_galerkin.py
```diff
@@ -42,17 +42,15 @@ def test_1d_heat_dirichletbc(order, basis_type):
...
         - inner(rhs, v) * dx
     )
-    bc = [
+    bcs = [
         DirichletBC(V, u_1, 2),
         DirichletBC(V, u_0, 1),
     ]
 
-    luparams = {"mat_type": "aij", "ksp_type": "preonly", "pc_type": "lu"}
+    sparams = {"snes_type": "ksponly", "ksp_type": "preonly", "pc_type": "lu"}
```
rckirby: If the test passes with default parameters, we can omit.

irksome/galerkin_stepper.py
```diff
@@ -142,35 +158,34 @@ def getFormGalerkin(F, L_trial, L_test, Qdefault, t, dt, u0, stages, bcs=None, a
...
     bcsnew = []
     for bc in bcs:
-        u0_sub = bc2space(bc, u0)
         g0 = as_ufl(bc._original_arg)
-        Vg_np = np.array([replace(g0, {t: t + c * dt}) for c in qpts])
-        Vg_np -= u0_sub * trial_vals[0]
-        g_np = proj @ Vg_np
+        dtg0 = expand_time_derivatives(TimeDerivative(g0), t=t, timedep_coeffs=(u0,))
         for i in range(num_stages):
+            # Evaluate the degrees of freedom
```
pbrubeck: I redid the BCs because the previous code relied on the default quadrature, which could potentially be under-integrating the L2 projection for Lobatto. These new BCs correspond to `bc_type="ODE"`, but we can easily restore the old ones as `bc_type="DAE"`

irksome/stage_derivative.py
```diff
@@ -75,20 +75,20 @@ def getForm(F, butch, t, dt, u0, stages, bcs=None, bc_type=None, splitting=AI):
...
+
     repl = {}
     for i in range(num_stages):
         repl[i] = {t: t + c[i] * dt,
                    v: v_np[i],
                    u0: u0 + as_tensor(A1w[i]) * dt,
                    dtu: A2invw[i]}
 
-    Fnew = sum(replace(F, repl[i]) for i in range(num_stages))
+    Fnew = sum(replace(F, repl(i)) for i in range(num_stages))
```
pbrubeck:
```suggestion
    Fnew = sum(replace(F, repl[i]) for i in range(num_stages))
```

tests/test_pc.py
```diff
@@ -29,7 +34,7 @@ def getNewForm(self, pc, u0, test):
         return F, bcs
 
 
-def rd(butcher_tableau):
+def rd(scheme, Lhigh=None, **kwargs):
```
rckirby: A comment for the reader (even though it's a test) that when `Lhigh` is nontrivial and you have a Galerkin collocation scheme, this approximates the Jacobian with a lower-order quadrature rule before applying Rana?

irksome/scheme.py
```diff
@@ -49,16 +49,49 @@ class ContinuousPetrovGalerkinScheme(GalerkinScheme):
...
         super().__init__(order, basis_type=basis_type,
                          quadrature_degree=quadrature_degree,
                          quadrature_scheme=quadrature_scheme)
 
 
-def create_time_quadrature(degree, scheme="default"):
+class GalerkinCollocationScheme(ContinuousPetrovGalerkinScheme):
+    """Class for describing collocation cPG-in-time methods"""
+    def __init__(self, order,
+                 stage_type="value",
```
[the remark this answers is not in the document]
pbrubeck: You are right "deriv" might be a better default, if we want to be consistent with the IRK default. But also the equivalent stage-value collocation IRK does not currently support `Dt(g(u))`

rckirby (approved):

PR #181, opened 2025-11-19
Simple measure wrapper allowing more easy specification of specific quadrature rules on individual terms

BorisAndrews:
This is similar to my [**old PR**](https://github.com/firedrakeproject/Irksome/pull/177), except interfaces now more naturally with the simpler interface for defining Galerkin in time schemes. :) I wasn't sure of good terminology; the names "`dx_override`" etc. can be placeholders.

Have included a simple nonlinear Schrodinger test, which should show how it works.


BorisAndrews:
That makes sense. Would you then say it's worth re-opening my old PR that leaned more into `TimeQuadratureLabel`, with some tests?

PR #183, opened 2025-12-02
TimeStepper: support auxiliary preconditioning form Fp

pbrubeck:
Enables passing an auxiliary semi-discrete Form `Fp` to construct a preconditioner for the stage-Jacobian. We obtain the preconditioner by differentiating the corresponding Form after time discretization.

This provides a much cleaner interface to auxiliary operator preconditioners, without requiring a custom `AuxiliaryOperatorPC` subclass:

```
TimeStepper(F, scheme, t, dt, u, Fp=Fp)
```

When `F` has `TimeQuadratureLabels`, under-integration in time can be achieved via `Fp = as_form(F)`.

Separate from quadrature-based techniques, one can also drop terms, use a simpler form, etc.

### DIFF
--- a/irksome/discontinuous_galerkin_stepper.py
+++ b/irksome/discontinuous_galerkin_stepper.py
@@ -218,8 +218,7 @@ class DiscontinuousGalerkinTimeStepper(StageCoupledTimeStepper):
         return getFormDiscGalerkin(F or self.F,
                                    el,
                                    quadrature or self.quadrature,
-                                   self.t, self.dt,
-                                   self.u0, stages, bcs)
+                                   self.t, self.dt, self.u0, stages, bcs)
 
     def _update(self):
         stages_np = np.array(self.stages.subfunctions, dtype=object)


tests/test_pc.py
```diff
@@ -203,3 +206,154 @@ def test_git_irk_equivalence(quad_scheme, order, stage_type):
...
+    invariants = [tuple(map(assemble, functionals))]
+    for _ in range(Nt):
+        stepper.advance()
+
+        invariants.append(tuple(map(assemble, functionals)))
+        i1, i2, i3 = invariants[-1]
+        t.assign(float(t) + float(dt))
+        times.append(float(t))
+
+        print(f'{float(t):.15f}, {i1:.15f}, {i2:.15f}, {i3:.15f}')
```
rckirby: Save printing for demos.  We need an `assert` or some such if this is actually a test.

tests/test_pc.py
```diff
@@ -203,3 +206,154 @@ def test_git_irk_equivalence(quad_scheme, order, stage_type):
         stepper.advance()
         assert numpy.allclose(stepper.solver.snes.ksp.computeEigenvalues(), 1.0)
         t.assign(float(t) + float(dt))
+
+
+@pytest.mark.parametrize('family', (ContinuousPetrovGalerkinScheme, RadauIIA))
```
rckirby: We need some kind of test criterion, that we get the correct answer or have a low iteration count or some such.

pbrubeck: I set `-ksp_max_it 15` and added tests for the eigenvalues

tests/test_pc.py
pbrubeck: I added `-ksp_max_it 5` and (very tight) assertions on the invariants

irksome/explicit_stepper.py
```diff
@@ -7,16 +7,15 @@
...
+            solver_parameters = {}
         solver_parameters["snes_lag_jacobian_persists"] = "true"
         solver_parameters["snes_lag_jacobian"] = -2
         solver_parameters["snes_lag_preconditioner_persists"] = "true"
         solver_parameters["snes_lag_preconditioner"] = -2
-        super(ExplicitTimeStepper, self).__init__(
-            F, butcher_tableau, t, dt, u0, bcs=bcs,
-            solver_parameters=solver_parameters, appctx=appctx,
-            nullspace=None)
+        super().__init__(F, butcher_tableau, t, dt, u0, bcs=bcs,
```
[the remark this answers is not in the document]
pbrubeck: These are handled by the DIRK superclass

irksome/dirk_stepper.py
```diff
@@ -191,9 +191,12 @@ def advance(self):
     def solver_stats(self):
         return self.num_steps, self.num_nonlinear_iterations, self.num_linear_iterations
 
-    def get_form_and_bcs(self, stages, tableau=None, F=None):
+    def get_form_and_bcs(self, stages, F=None, u0=None, bcs=None, tableau=None):
```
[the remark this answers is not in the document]
pbrubeck: I'm happy to leave `u0` as a kwarg but can also delete it if you really don't like it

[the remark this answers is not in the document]
rckirby: I really think it should go away until we know we need it and how it should be used.  Also, untested therefore broken.  It could be (mis)used and produce results we don't understand.

irksome/dirk_stepper.py
```diff
@@ -191,9 +199,12 @@ def advance(self):
...
+        if bcs is None:
+            bcs = self.orig_bcs
         return getFormDIRK(F or self.F,
                            stages,
                            tableau or self.butcher_tableau,
                            self.t, self.dt,
-                           self.u0, bcs=self.orig_bcs)
+                           u0 or self.u0,
```
```suggestion
                           self.u0,
```

irksome/galerkin_stepper.py
```diff
@@ -315,12 +315,16 @@ def __init__(self, F, scheme, t, dt, u0, bcs=None,
...
         self.set_update_expressions()
 
-    def get_form_and_bcs(self, stages, tableau=None, basis_type=None, order=None, quadrature=None, aux_indices=None, F=None):
+    def get_form_and_bcs(self, stages, F=None, u0=None, bcs=None,
+                         tableau=None, basis_type=None, order=None,
+                         quadrature=None, aux_indices=None):
         F = F or self.F
-        bcs = self.orig_bcs
-        aux_indices = aux_indices or self.aux_indices
+        u0 = u0 or self.u0
```
```suggestion
```

irksome/stage_derivative.py
```diff
@@ -223,12 +224,16 @@ def _update(self):
         for i, u0bit in enumerate(self.u0.subfunctions):
             u0bit += sum(self.stages.subfunctions[nf * s + i] * (b[s] * dt) for s in range(ns))
 
-    def get_form_and_bcs(self, stages, tableau=None, F=None):
+    def get_form_and_bcs(self, stages, F=None, u0=None, bcs=None, tableau=None):
```
pbrubeck:
```suggestion
    def get_form_and_bcs(self, stages, F=None, bcs=None, tableau=None):
```

irksome/galerkin_stepper.py
```diff
@@ -366,7 +370,7 @@ def scaledIA(A):
             trial_el, test_el = getElements(basis_type, order)
         quadrature = quadrature or self.quadrature
         return getFormGalerkin(F, trial_el, test_el, quadrature,
-                               self.t, self.dt, self.u0, stages,
+                               self.t, self.dt, u0, stages,
```
```suggestion
                               self.t, self.dt, self.u0, stages,
```

irksome/base_time_stepper.py
```diff
@@ -115,7 +121,7 @@ def __init__(self, F, t, dt, u0, num_stages,
         self.bigBCs = bigBCs
 
         self.problem = NonlinearVariationalProblem(
-            Fbig, stages, bcs=bigBCs,
+            Fbig, stages, bcs=bigBCs, J=Jbig, Jp=Jpbig,
```
pbrubeck:
```suggestion
            Fbig, stages, bcs=bigBCs, Jp=Jpbig,
```

PR #191, opened 2026-01-29
Isolating UFL manipulation into ufl/ submodule

436ahsan:
This is a draft pull request to begin isolating UFL manipulation routines into a dedicated `irksome/ufl/`submodule, as discussed previously. The goal is to eventually share ufl between Firedrake and a future FEniCS-X backend. 

This is not fully complete, just a first step. I am opening it early to gather feedback from @rckirby and @pbrubeck

Happy to revise based on suggestions.

### DIFF
--- a/irksome/backend.py
+++ b/irksome/backend.py
@@ -2,9 +2,9 @@ from typing import Protocol
 import ufl
 from importlib import import_module
 
-class Backend(Protocol):
 
-    def function_space(self, V: ufl.Coefficient) -> ufl.FunctionSpace:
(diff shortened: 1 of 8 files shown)


irksome/ufl/__init__.py
```diff
@@ -0,0 +1 @@
+from .manipulation import SplitTimeForm, check_integrals, extract_terms, strip_dt_form
```
pbrubeck: This wasn't part of the public API before, we should not expose it now as public

PR #199, opened 2026-03-02
Jdstephens/multistep

jdsteph2:
Initial work on automating general multistep methods.


irksome/multistep.py
```diff
@@ -0,0 +1,185 @@
...
+            to find the required starting values.
+    """
+
+    def __init__(self, F, method, t, dt, u0, bcs=None, Fp=None, solver_parameters=None, bounds=None, appctx=None, nullspace=None,
+                 transpose_nullspace=None, near_nullspace=None, startup_parameters=None, **kwargs):
+
+        super().__init__(F, t, dt, u0,
+                         bcs=bcs, appctx=appctx, nullspace=nullspace)
+
+        assert isinstance(method, MultistepTableau)
```
rckirby: Assert before you call base class constructor?

rckirby (commented):
First quick scan looks mostly good.
I have a few comments above.
But we should also add a demo, especially showing the automated starting scheme with RK?

irksome/multistep.py
```diff
@@ -0,0 +1,185 @@
...
+        F_dt = split_form.time
+        F_remainder = split_form.remainder
+
+        # replace the time derivative with a linear combination of the previous steps
+        temp_form = 0.0
+        for (i, coeff) in enumerate(a):
+            if coeff is zero():
+                pass
+            else:
+                temp_form += coeff * self.us[i]
```
pbrubeck: UFL can handle zero cancellation quite well
```suggestion
            temp_form += coeff * self.us[i]
```

rckirby (commented):
Can we update `TimeStepper` to return a `MultistepTimeStepper` if you pass in BDF, etc?

demos/heat/demo_heat_multistep.py.rst
```diff
@@ -0,0 +1,136 @@
...
+  startup_stepper_kwargs = {'stage_type': 'value', 
+                            'solver_parameters': luparams}
+
+  startup_parameters = {'tableau': RadauIIA(1),
+                        'dt_div': 8,
+                        'stepper_kwargs': startup_stepper_kwargs}
+
+We can then set up the :class:`.MultistepTimeStepper` as follows::
+
+  stepper = MultistepTimeStepper(F, method, t, dt, u, bcs=bc,
```
rckirby: Ok, you should just be able to add a kwarg to `TimeStepper` (that only works if you provide a `MultistepTimeStepper` as the `scheme` called `multistep_startup_parameters`.  Then, that dict just gets passed on to the `MultistepTimeStepper` when you return it from `TimeStepper`.

rckirby (changes requested):
Nice demo, I have a few comments on it.

demos/heat/demo_heat_multistep.py.rst
jdsteph2: I'll give this another go. As far as I can tell, this will necessarily lead to a circular import because the startup procedure creates an instance of a `TimeStepper`.

rckirby: Not sure, don't we do this for all the other steppers?  Check out the stage value stepper, etc

rckirby (changes requested):

rckirby (approved):
I think we're up to date now and ready to go.  @pbrubeck and @ScottMacLachlan?

demos/heat/demo_heat_multistep.py.rst
```diff
@@ -0,0 +1,155 @@
...
+   u_t - \Delta u &= f
+
+   u & = 0 \quad \textrm{on}\ \Gamma
+
+for some known function :math:`f`.  At each time :math:`t`, the solution
+to this equation will be some function :math:`u\in V`, for a suitable function
+space :math:`V`.
+
+We transform this into weak form by multiplying by an arbitrary test function
+:math:`v\in V` and integrating over :math:`\Omega`.  We know have the
```
ScottMacLachlan: `know` -> `now`

demos/heat/demo_heat_multistep.py.rst
```diff
@@ -0,0 +1,155 @@
...
+This demo implements an example used by Solin with a particular choice
+of :math:`f` given below
+
+As usual, we need to import firedrake::
+
+  from firedrake import *
+
+We will also need to import certain items from irksome::
+
+  from irksome import Dt, MeshConstant, TimeStepper, MultistepMethod
```
ScottMacLachlan: @pbrubeck will (rightly) point out there's no need for `MeshConstant` in a demo.

demos/heat/demo_heat_multistep.py.rst
```diff
@@ -0,0 +1,155 @@
...
+On the other hand, if you wish to use a method which Irksome supports to obtain these starting values, then Irksome allows this process to be completed automatically.
+
+We'll showcase both methods. The automatic process requires defining a :class:`dict` of keyword arguments used to setup a :class:`.TimeStepper` representing a single-step method. This :class:`.TimeStepper` is then used to compute the 
+initial approximations needed to start the method. We'll import RadauIIA and use the backward Euler method to obtain our starting values. Formally, the backward Euler method is only first order accurate, 
+and we wish to use it obtain the starting values for the third-order accurate BDF(3) method. A crude way to increase the accuracy of the starting values is to use a smaller timestep for the startup procedure. 
+Here, we use timesteps of size :math:`\Delta t / 8` for the backward Euler method which is accessible through the keyword `dt_div`. The keyword `stepper_kwargs` allows for easy customization of the startup :class:`.TimeStepper`.::
```
ScottMacLachlan: Would be nice to be clear that `dt_div = 8` gives 8 startup time steps per stepper timestep.  (I.e., define `dt_div` explicitly here)

demos/heat/demo_heat_multistep.py.rst
```diff
@@ -0,0 +1,155 @@
...
+                        'stepper_kwargs': startup_stepper_kwargs}
+
+We can then set up the :class:`.TimeStepper` as follows::
+
+  stepper = TimeStepper(F, method, t, dt, u, bcs=bc, 
+                                 solver_parameters=luparams, 
+                                 startup_parameters=startup_parameters)
+
+Note that the creation of a :class:`.TimeStepper` configured for a multistep method with parameters for the startup procedure will not automatically solve for the required starting values. 
+One must call the :class:`.TimeSteppers`'s :meth:`~.TimeStepper.startup` method, which will internally construct the single-step :class:`.TimeStepper`, solve the the required starting
```
ScottMacLachlan: `the the` -> `the`

irksome/multistep.py
```diff
@@ -0,0 +1,173 @@
...
+                assert butcher_tableau.num_total_steps == 2, "Cannot use a multistep method to start a multistep method"
+            stepper_kwargs = self.startup_parameters.get('stepper_kwargs', {})
+            startup_dt_div = self.startup_parameters.get('dt_div', 1)
+            assert isinstance(startup_dt_div, int) and startup_dt_div > 0
+
+            # delayed import to avoid a circular import
+            from .stepper import TimeStepper
+
+            self.us[0].assign(self.u0)
+            self.TS = TimeStepper(self.F, butcher_tableau, self.t, self.dt, self.u0, bcs=self.bcs, **stepper_kwargs)
```
ScottMacLachlan: Needs a better name.  `self.startup_stepper` or similar.

irksome/multistep.py
```diff
@@ -0,0 +1,173 @@
...
+
+            # modify the timestep
+            self.dt.assign(self.dt / startup_dt_div)
```
ScottMacLachlan: Why are we modifying `self.dt` here, when we just want a smaller timestep for the startup stepper?  We shouldn't touch `self.dt` at all in this process.

irksome/multistep.py
```diff
@@ -0,0 +1,173 @@
...
+
+            # advance the system and assign values to previous steps
+            for i in range(self.num_prev_steps - 1):
+                for substep in range(startup_dt_div):
+                    self.TS.advance()
+                    self.t.assign(self.t + self.dt)
```
ScottMacLachlan: Again, I would not touch `self.t` in the code

irksome/multistep.py
ScottMacLachlan: Create a new `Constant` or `MeshConstant` for the timestep for the startup stepper

irksome/multistep.py
```diff
@@ -0,0 +1,173 @@
...
+                self.us[i + 1].assign(self.u0)
+            self.dt.assign(self.dt * startup_dt_div)
```
ScottMacLachlan: No.  Just.  No.

demos/heat/demo_heat_multistep.py.rst
```diff
@@ -0,0 +1,136 @@
...
+We can then set up the :class:`.MultistepTimeStepper` as follows::
+
+  stepper = MultistepTimeStepper(F, method, t, dt, u, bcs=bc, 
+                                 solver_parameters=luparams, 
+                                 startup_parameters=startup_parameters)
+
+Note that the creation of a :class:`.MultistepTimeStepper` with parameters for the startup procedure will automatically solve for the required starting values and advance `t` to `t + (s-1)*dt`. This is not the case 
+if startup parameters are not provided. In that case, the user is responsible for setting the initial values and advancing `t` to the appropriate value. In our case, we see that time has been advances:::
+
+  print(f'The starting values have been computed. The current time is {float(t)}')
```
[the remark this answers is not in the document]
ScottMacLachlan: The code in `.startup()` still advances `t`.  I would remove that and document here that it is the user's responsibility.

demos/heat/demo_heat_multistep.py.rst
```diff
@@ -0,0 +1,155 @@
...
+                                 startup_parameters=startup_parameters)
+
+Note that the creation of a :class:`.TimeStepper` configured for a multistep method with parameters for the startup procedure will not automatically solve for the required starting values. 
+One must call the :class:`.TimeSteppers`'s :meth:`~.TimeStepper.startup` method, which will internally construct the single-step :class:`.TimeStepper`, solve the the required starting 
+values, and advance `t` to `t + (s-1)*dt`:::
+
+  stepper.startup()
+  print(f'The starting values have been computed. The current time is {float(t)}')
+
+If you would rather set the starting values by hand (for instance, in this case the exact solution is available) the process is straightforward. We'll first reset `t` so that everything is consistent, and then begin projecting the exact solution
```
ScottMacLachlan: `in this case` -> `in case`

ScottMacLachlan: Is there a reason to `project` and not `interpolate` here?

irksome/multistep.py
```diff
@@ -0,0 +1,173 @@
...
+        F(t, u; v) == 0, where `u` is the unknown
+        :class:`firedrake.Function and `v` is the
+        :class:firedrake.TestFunction`.
+    :arg method: A :class:`MultistepMethod` corresponding to the desired multistep method.
+    :arg t: a :class:`Function` on the Real space over the same mesh as
+         `u0`.  This serves as a variable referring to the current time.
+    :arg dt: a :class:`Function` on the Real space over the same mesh as
+         `u0`.  This serves as a variable referring to the current time step.
+         The user may adjust this value between time steps.
+    :arg u: A :class:`firedrake.Function` containing the current
```
ScottMacLachlan: This seems to be referred to as `u0` in the previous line, and in the `__init__` method.  Why `u` here?

irksome/multistep.py
```diff
@@ -0,0 +1,173 @@
...
+        F_dt = split_form.time
+        F_remainder = split_form.remainder
+
+        # replace the time derivative with a linear combination of the previous steps
+        temp_form = 0.0
+        for (i, coeff) in enumerate(a):
+            temp_form += coeff * self.us[i]
+
+        dtu = TimeDerivative(u0)
+        Fnew = replace(F_dt, {dtu: temp_form})
```
ScottMacLachlan: Does this DTRT for all nonlinear problems?  (Or all nonlinear problems that we care about?)

ScottMacLachlan: Put another way: do we need a disclaimer on what problems this can handle?

tests/test_multistep.py
```diff
@@ -0,0 +1,622 @@
...
+
+    for i in range(5):
+
+        stepper.advance()
+        t.assign(float(t) + float(dt))
+
+    return u
+
+
+def heat(msh, N, spatial_degree):
```
ScottMacLachlan: Needs a better name, `heat_BDF2_hand`?

tests/test_multistep.py
```diff
@@ -0,0 +1,622 @@
...
+    for i in range(5):
+        t.assign(float(t) + float(dt))
+        stepper_solver.solve()
+        u0.assign(u1)
+        u1.assign(u2)
+
+    return u2
+
+
+def heat_mech(msh, N, spatial_degree):
```
ScottMacLachlan: Needs a better name, `heat_BDF2_mech`?

tests/test_multistep.py
```diff
@@ -0,0 +1,622 @@
...
+    stepper.us[1].assign(u1)
+
+    for i in range(5):
+        stepper.advance()
+        t.assign(float(t) + float(dt))
+
+    return u
+
+
+def heat_bounds(bounds_flag, startup_bounds_flag, startup_tableau):
```
ScottMacLachlan: `heat_BDF2_bounds`?

tests/test_multistep.py
```diff
@@ -0,0 +1,622 @@
...
+    min_init = min(stepper.us[0].dat.data)
+    min_step1 = min(stepper.us[1].dat.data)
+    stepper.advance()
+    t.assign(t + dt)
+    min_step2 = min(u.dat.data)
+
+    return (min_init >= 0.0, min_step1 >= 0.0, min_step2 >= 0)
+
+
+def CH_hand(msh, spatial_degree, startup_tableau):
```
ScottMacLachlan: Needs a better name.  Also, it isn't clear what this is testing that the heat tests aren't?  I guess mixed/nonlinear?

tests/test_multistep.py
```diff
@@ -0,0 +1,622 @@
...
+
+    stepper = TimeStepper(F, method, t, dt, u, bcs=bc, startup_parameters=startup_parameters)
+    stepper.startup()
+
+    for i in range(10):
+        stepper.advance()
+        t.assign(float(t) + float(dt))
+
+    return u
+
```
ScottMacLachlan: There's a shocking amount of repeated code in these tests.  I appreciate the thoroughness, but it feels slightly unmaintainable.  Is it possible to simplify, so we can avoid having to touch too many pieces of code to update if syntax changes?

ScottMacLachlan (changes requested):
Looks like this is getting close, but I've got some concerns.  The biggest ones are in the `.startup()` routine, which feels messier than needed (we step on ourselves in a few spots).  These get reflected in the demo.  And, then, there's a lot going on in those tests.  It feels like we could do this more elegantly, rather than simply copy and paste the same heat equation example umpteen times...

irksome/multistep.py
jdsteph2: Done. Doing this requires grabbing a new copy of the semidiscrete form and the boundary conditions which use the new `t`. I have something in that's working right now, but there might be a better way to handle this?

jdsteph2:
> Looks like this is getting close, but I've got some concerns. The biggest ones are in the `.startup()` routine, which feels messier than needed (we step on ourselves in a few spots). These get reflected in the demo. And, then, there's a lot going on in those tests. It feels like we could do this more elegantly, rather than simply copy and paste the same heat equation example umpteen times...

Copy all.  Thank you for the feedback!  I still have to make some of the smaller edits and pare down the tests. 

I've removed the unpleasant bits of the startup procedure (those that adjusted `t` and `dt`), but this presented a new problem. It appears that we will have to grab a new copy of the semidiscrete form and the boundary conditions which reference the newly defined `startup_t`. I have something in that is working right now, but there may be a cleaner way to handle this?

demos/heat/demo_heat_multistep.py.rst
pbrubeck: Maybe we could come up with a better name, perhaps `num_startup_steps`?

irksome/tableaux/multistep_tableaux.py
```diff
@@ -0,0 +1,120 @@
...
+    @property
+    def is_explicit(self):
+        return np.abs(self.b[-1]) < 1e-10
+
+    @property
+    def is_implicit(self):
+        return not self.is_explicit
+
+
+class MultistepMethod(MultistepTableau):
```
pbrubeck: Why are there two classes? Also could we not use a different name, such as `MultistepScheme` to align with `ContinuousPetrovGalerkinScheme` and `DGScheme`?

Alternatively we can keep `MultistageTableau,` which aligns with `ButcherTableau`. I'm suggesting this to minimize the terminology used to specify the time stepper scheme.

irksome/tableaux/multistep_tableaux.py
```diff
@@ -0,0 +1,120 @@
...
+            a, b = get_weights_AB(order)
+        elif method == 'AM':
+            a, b = get_weights_AM(order)
+        else:
+            a, b = get_weights_BDF(order)
+
+        super().__init__(a, b)
+
+
+def get_weights_BDF(order):
```
pbrubeck, line 55: We have implemented the different Butcher tableaux as subclasses. Should we not follow the same pattern for `MultistageTableau`?

One advantage is that users will be to referring to the scheme with its full name. Currently one can only select a method from the initials and this might not be helpful for someone who is not familiar with all the options.

irksome/tableaux/multistep_tableaux.py
jdsteph2: I am indifferent between scheme, tableau, and method. Method seems to be the most standard for multistep, but any should work.

irksome/tableaux/multistep_tableaux.py
jdsteph2: Agreed.

demos/heat/demo_heat_multistep.py.rst
jdsteph2: Ah, fair point.

demos/heat/demo_heat_multistep.py.rst
jdsteph2: What is the current status of MeshConstant?  I believe I just copied this from one of the other demos.

rckirby: Just use Constant.  MeshConstant didn't solve the problem we thought it would, which has had another solution.  We should do a pervasive fix at some point, but don't add work for that in new development.

irksome/multistep.py
jdsteph2: Ah, turns out we already do this elsewhere.

demos/heat/demo_heat_multistep.py.rst
jdsteph2: Got it. Thanks!

irksome/multistep.py
rckirby: This is almost surely broken for `Dt(g(u))`.  I would expect the right thing for BDF for, say, Richards equation would be to replace it with `g(u^{n+1}) - g(u^n)` or more generally, the sum of `u^{n-j}` times appropriate scalars.  Probably looking at the somewhat similar UFL-fu in stage value would be helpful.

I pass on whether this replacement is the right thing for Adams-type methods, but if you're using those for Richards equation then you have deeper problems than Irksome possibly being wrong ;)

jdsteph2: I think I have the replacement for problems with Dt(g(u)) working properly now.  What I have at least passes the Richards equation mass conservation test for BDF(1) and AM(0) - AM(2).

BDF(2) fails the test we have in place for the Richards equation...not sure if its expected to pass in theory or not.

irksome/multistep.py
```diff
@@ -0,0 +1,188 @@
...
+from .ufl.deriv import expand_time_derivatives
+from .base_time_stepper import BaseTimeStepper
+from .tableaux.multistep_tableaux import MultistepTableau
+from .bcs import stage2spaces4bc
+from ufl import Form
+from ufl.constantvalue import as_ufl
+from firedrake import NonlinearVariationalProblem, NonlinearVariationalSolver, derivative, Constant
+
+
+class MultistepTimeStepper(BaseTimeStepper):
```
pbrubeck, line 13: There is a lot of code duplication, it seems that this class wants to inherit from StageCoupledTimeStepper.

jdsteph2: Multistep methods do not require the creation of a larger function space. The previous values are only used as data in a variational problem posed over `u0.function_space()`. I suppose this could be done, but I'm not sure it is worth the trouble?

pbrubeck: Ah, I didn't see that. Maybe num_stages = 1 does the trick?

Then this looks more like a DIRK. We do not have a base class for those, so the best plan is to leave that work for a future PR.

jdsteph2: I think you could force it into a StageCoupledTimeStepper with num_stages = 1, but it will get pretty messy.  Copy--I'll leave it as is for now.

irksome/multistep.py
jdsteph2: > BDF(2) fails the test we have in place for the Richards equation...not sure if its expected to pass in theory or not.

It looks like Dt(g(u)) should be discretized as `a_s g(u^{n + s}) + ... a_0 g(u^{n})`. This matches what we have for stage-value, and leads to mass conservation for the Adams-Moulton methods (both in theory and in practice).

Discretizing in this way, BDF(2) and beyond will not conserve mass because the conservative finite difference does not appear in the form. You instead get something that depends on the mass flux and the previous values. This appears to be a limitation of the BDF family for such problems, and not the discretization method.

irksome/multistep.py
rckirby: Trying to hook multistage into DIRK or stage-coupled is silly.  There is actually a unifying mathematical framework that encompasses both RK and multistep -- these are General Linear Methods.

Doing a full push on GLM could unify these  methods (but we probably still need special cases for stage-sequential things like explicit/DIRK and Galerkin doesn't wedge in cleanly).  Not sure what we'd gain code and if there are methods here that we *ought* to be using but aren't because they are too complicated to implement...

At any rate, I concur with leaving this alone.

demos/heat/demo_heat_multistep.py.rst
```diff
@@ -0,0 +1,147 @@
...
+
+We can then set up the :class:`.TimeStepper` as follows::
+
+  stepper = TimeStepper(F, method, t, dt, u, bcs=bc, 
+                                 solver_parameters=luparams, 
+                                 startup_parameters=startup_parameters)
+
+Note that the creation of a :class:`.TimeStepper` configured for a multistep method with parameters for the startup procedure will not automatically solve for the required starting values. 
+One must call the :class:`.TimeSteppers`'s :meth:`~.TimeStepper.startup` method, which will internally construct the single-step :class:`.TimeStepper` and solve for the required starting 
+values. It is the users responsibility to advance `t` to `t + (s-1)*dt`. Unless more refined control is needed, this value is available as :meth:`.TimeStepper.startup_t`::
```
ScottMacLachlan: `users` -> `user's`

ScottMacLachlan (approved):
Aside from one typo, this looks great!  Many thanks!

demos/heat/demo_heat_multistep.py.rst
rckirby:
```suggestion
values. It is the user's responsibility to advance `t` to `t + (s-1)*dt`. Unless more refined control is needed, this value is available as :meth:`.TimeStepper.startup_t`::
```

rckirby: Thanks, I took care of this.

jdsteph2:
Thank you all so much for your help!

PR #202, opened 2026-03-24
Add convenience function to sample functions over timestep

ScottMacLachlan:
The main Irksome interface keeps track of the solution at the time steps, but doesn't provide a mechanism to sample at time points within a timestep.  This provides a mechanism to specify points where we want to generate sample values (e.g., for quadrature to evaluate time integrals).

Full credit goes to @jdsteph2 for doing all the hard work here - many thanks!


rckirby (approved):
Thanks, I think this looks good.  The more general family of things in the RK literature goes by "dense output".  We're not supporting those with this, but future pull requests are welcome ;)

tests/test_collocation_eval.py
```diff
@@ -0,0 +1,213 @@
...
+    qs = [u_interp.copy(deepcopy=True)]
+    ts.append(float(t))
+
+    for i in range(2):
+
+        stepper.advance()
+
+        for (i, val) in enumerate(stepper.sample_values):
+            ts.append(float(t) + sample_points[i] * float(dt))
+            u_interp.interpolate(val)
```
pbrubeck: What are we expecting `stepper.sample_values` to be? Should it be a list `Functions` in same space as `stepper.u0`? It feels wrong to interpolate here

irksome/tools.py
```diff
@@ -32,6 +32,18 @@ def get_stage_space(V, num_stages):
...
 
+def split_stages(V, stages):
+    """Reconstruct the stages as a list of Function(V)"""
+    num_fields = len(V)
+    if num_fields == 1:
+        return stages.subfunctions
+
+    dats = stages.dat
+    ks = [Function(V, val=MixedDat(dats[offset:offset+num_fields]))
+          for offset in range(0, len(dats), num_fields)]
```
pbrubeck:
```suggestion
    stages_np = reshape(stages, (-1, *V.value_shape))
    ks = [stages_np[i] for i in range(stages_np.shape[0])]
```

tests/test_collocation_eval.py
```diff
@@ -0,0 +1,224 @@
...
+    e_vals = []
+    check = Function(qu)
+    for step in range(2):
+
+        stepper.advance()
+
+        for (i, val) in enumerate(stepper.sample_values):
+            my_t.assign(float(t) + sample_points[i] * float(dt))
+            check.assign(val)
+            e_vals.append(errornorm(my_qu, check))
```
pbrubeck:
```suggestion
            check.interpolate(val)
            e_vals.append(errornorm(my_qu, check))
```

irksome/tools.py
```diff
@@ -1,7 +1,7 @@
 from operator import mul
 from functools import reduce
 import numpy
-from firedrake import VectorSpaceBasis, MixedVectorSpaceBasis
+from firedrake import Function, VectorSpaceBasis, MixedVectorSpaceBasis
```
pbrubeck:
```suggestion
from firedrake import VectorSpaceBasis, MixedVectorSpaceBasis
```

irksome/base_time_stepper.py
```diff
@@ -228,6 +239,27 @@ def get_stage_bounds(self, bounds=None):
...
+        '''
+        pts = numpy.reshape(self.sample_points, (-1, 1))
+        vander = self.tabulate_poly(pts)
+
+        self.u_old = Function(self.u0)
+        ks = [self.u_old]
+        ks.extend(split_stages(self.u0.function_space(), self.stages))
+        num_samples = vander.shape[1]
+        self.sample_values = [sum(ks[j] * vander[j, i] for j in range(len(ks)))
+                              for i in range(num_samples)]
```
pbrubeck, line 262:
```suggestion
                              for i in range(num_samples)]

```

rckirby (approved):
I'm happy with this.  Thanks for all the legwork on the messy mixed business!

PR #204, opened 2026-04-01
Conservative StageValueStepper

pbrubeck:
Currently, we splat out time derivatives of `Dt(g(u))` in equations eagerly, but this is inconsistent with how collocation schemes are formulated.  Instead, we want to keep this term together, which can help with mass conservation for things like Richards equation.  A similar feature appears in DG time stepping, where the derivation requires us to work on `Dt(g(u))` itself to IBP twice.  

This PR only addresses stage-value, but we can adapt some of the UFL manipulation to DG later.

It seems that stage derivative is essentially hopeless here, there's no way to do substitution on `Dt(g(u))` without splatting out the derivatives.

cPG may be salvagable, but that's an open question.

### DIFF
--- a/irksome/discontinuous_galerkin_stepper.py
+++ b/irksome/discontinuous_galerkin_stepper.py
@@ -50,10 +50,11 @@ def getTermDiscGalerkin(F, L, Q, t, dt, u0, stages, test):
     basis_vals = tabulate_basis[(0,)]
     basis_dvals = tabulate_basis[(1,)]
     basis_vals_w = np.multiply(basis_vals, qwts)
+    basis_dvals_w = np.multiply(basis_dvals, qwts)
 
     trial_vals = vecconst(basis_vals)
-    trial_dvals = vecconst(basis_dvals)
     test_vals_w = vecconst(basis_vals_w)
+    test_dvals_w = vecconst(basis_dvals_w)
     qpts = vecconst(qpts.reshape((-1,)))
 
     # set up the pieces we need to work with to do our substitutions
@@ -61,7 +62,7 @@ def getTermDiscGalerkin(F, L, Q, t, dt, u0, stages, test):
     u_np = reshape(stages, (-1, *u0.ufl_shape))
     vsub = dot(test_vals_w.T, v_np)
     usub = dot(trial_vals.T, u_np)
-    dtu0sub = dot(trial_dvals.T, u_np)
+    dtvsub = dot(test_dvals_w.T, v_np)
 
     # preprocess time derivatives
     split_form = split_time_derivative_terms(F, t=t, timedep_coeffs=(u0,))
@@ -70,20 +71,23 @@ def getTermDiscGalerkin(F, L, Q, t, dt, u0, stages, test):
     if F_dtless.empty():
         Fnew = F_dtless
     else:
-        # Jump terms
-        L_at_0 = vecconst(L.tabulate(0, (0.0,))[(0,)])
-        u_at_0 = L_at_0 @ u_np
-        v_at_0 = L_at_0 @ v_np
-        repl = {u0: u_at_0 - u0,
-                v: v_at_0}
-        Fnew = replace(F_dtless, repl)
+        # Integrate by parts in time (Dt(g(u)), v)
+        # Jump terms: [(g(u), v)](t+dt) - [(g(u), v)](t)
+        ref_el = L.get_reference_element()
+        L_at_01 = vecconst(L.tabulate(0, ref_el.vertices)[(0,)])
+        u_at_01 = dot(L_at_01.T, u_np)
+        v_at_01 = dot(L_at_01.T, v_np)
 
-        # Terms with time derivatives
+        repl_old = {v: v_at_01[0]}
+        repl_new = {v: v_at_01[1], u0: u_at_01[1], t: t + dt}
+        Fnew = replace(F_dtless, repl_new) - replace(F_dtless, repl_old)
+
+        # Terms with time derivatives: -(g(u), Dt(v))
         for q in range(len(qpts)):
             repl = {t: t + qpts[q] * dt,
-                    v: vsub[q] * dt,
-                    u0: dtu0sub[q] / dt}
-            Fnew += replace(F_dtless, repl)
+                    v: dtvsub[q],
+                    u0: usub[q]}
+            Fnew -= replace(F_dtless, repl)
 
     # Handle the rest of the terms
     for q in range(len(qpts)):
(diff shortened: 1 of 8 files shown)


irksome/discontinuous_galerkin_stepper.py
```diff
@@ -66,11 +63,13 @@ def getTermDiscGalerkin(F, L, Q, t, dt, u0, stages, test):
...
-        F_dtless = strip_dt_form(split_form.time)
-        F_remainder = split_form.remainder
-
+    # preprocess time derivatives
+    split_form = split_time_derivative_terms(F, t=t, timedep_coeffs=(u0,))
+    F_dtless = remove_time_derivatives(split_form.time)
+    F_remainder = expand_time_derivatives(split_form.remainder, t=t, timedep_coeffs=())
+    if F_dtless.empty():
+        Fnew = F_dtless
+    else:
```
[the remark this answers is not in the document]
rckirby: If so, should we be adding that to the parameters for the mass conservation test?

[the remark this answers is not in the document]
rckirby: Probably something is needed on the white board?  Maybe we should land this without the fix for DG and come back when we know the right thing to do -- can we add an error message for `Dt(g(u[i]))` rather than `Dt(u[i])` and maybe a suggestion to try out stage value RK in the meantime?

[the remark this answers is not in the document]
pbrubeck: The tests pass, so what we have in the PR seems more correct than whatever we had before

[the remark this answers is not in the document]
pbrubeck: Integrating by parts twice forces a `Dt(g(u))` to appear in the substitution. However, we only know how to substitute `Dt(u)`, and this was fine when we splatted things out. By integrating by parts only once, the `Dt` is transferred to the `TestFunction`.

irksome/stage_value.py
```diff
@@ -73,8 +73,6 @@ def getFormStage(F, butch, t, dt, u0, stages, bcs=None, splitting=AI, vandermond
        - `bcnew`, a list of :class:`firedrake.DirichletBC` objects to be posed
          on the stages
     """
-    # preprocess time derivatives
-    F = expand_time_derivatives(F, t=t, timedep_coeffs=(u0,))
     # we can only do DAE-type problems correctly if one assumes a stiffly-accurate method.
     assert is_ode(F, u0) or butch.is_stiffly_accurate
```
pbrubeck, line 79: The test added in #205 now breaks, since `expand_time_derivatives` forces `Dt(u[i]) -> Dt(u)[i]`, hence every component of u will appear under `Dt`. Should we remove the failing case or should we remove this check?

I think `is_ode` needs to be much more sophisticated than what we have now in order to check e.g. `Dt(g(u[i]))`. Maybe we don't need this check after #205. Thoughts @ScottMacLachlan?

ScottMacLachlan: `test_dae.py::test_foo` is an old test, not part of the tests added in #205.

ScottMacLachlan: It looks to me like `is_ode` is only used in `getFormStage` in `stage_value.py`, but the comment that precedes it is no longer true with #205.  So, I'd be in favour of getting rid of it.  What does @rckirby think?

ScottMacLachlan: (Also, this must have been broken earlier for me not to have tripped over it when I started on the mixed heat test in #202...)

pbrubeck: The commit for which the test in #205 was breaking is [the merge commit](https://github.com/firedrakeproject/Irksome/pull/204/commits/e6263ccf6b92cbde662247ab78b9f800628ed5ab) and here's the [CI failure](
https://github.com/firedrakeproject/Irksome/actions/runs/24144015793/job/70452964544).

I then tried to fix `is_ode` by expanding derivatives within, but that broke `test_foo`. Maybe we should get rid of `test_foo` as well.

rckirby: Let me spell out the logic/features for stage value to work.  The old logic is intertwining things about the problem and the method:

- The derivation of stage value form requires each equation in the system to have time derivatives enter only in the form of `inner(Dt(g(u_i)), v) * dx(some_region)`.  However, we can have bits of the overall `u` appearing.  I believe the math works also (not sure about our substitution) if one has `inner(Dt(g(u1) + g(u2), v)` in one equation.  As soon as we decide not to support this, there will be an issue raised from some prospective user (and we have a version of this in Nystrom in our poroelastic example)
- Previously, if we were doing stage value on a DAE (e.g. Stokes or mixed Poisson) where some equations lacked a time derivative, we needed a stiffly accurate method so we could do "last = next" for the update, as our mass solve didn't do the right thing.  There is now the intermediate option of not stiffly accurate but invertible Butcher matrix.

So, we at least need to document the expected form (what we support) if we can't detect/fail.  Given the generalization to supporting `Dt(g(u))` and Scott's updated logic, it seems clear that the `is_ode or is_stiffly_accurate` logic is incorrect.

ScottMacLachlan: Yes - I agree with Rob, that the "old logic" is wrong.  There's nothing wrong with stage value for DAEs, so long as the BT is invertible.  (You can see this by the fact that we have an invertible map between the stage derivatives and stage values in this setting, so if stage derivative is well-posed, then stage value must be as well.)

ScottMacLachlan: So, the change to `is_ode` in this PR is fixing something that was broken before, otherwise we would have hit this in #202.  Either way, I think the safe fix is to add `det(A) !=0` to the logic here, since that's a case that is guaranteed(?) to work

pbrubeck: I don't trust `is_ode`, it'd fail to detect an ODE with `Dt(g(u))`. We can add `det(A) != 0`, and raise a warning. I'm happier if the check goes away.

irksome/discontinuous_galerkin_stepper.py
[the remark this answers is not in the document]
rckirby: Ok, we now have options for weak and strong DG forms that work, so I can close this bit of the conversation.

irksome/discontinuous_galerkin_stepper.py
```diff
@@ -116,6 +136,7 @@ def getFormDiscGalerkin(F, L, Qdefault, t, dt, u0, stages, bcs=None):
     :arg bcs: optionally, a :class:`DirichletBC` object (or iterable thereof)
          containing (possibly time-dependent) boundary conditions imposed
          on the system.
+    :arg deriv_type: either `"weak"` or `"strong"`.
```
pbrubeck, line 139: We need to say more here

irksome/scheme.py
pbrubeck, line 36: This needs a proper docstring

PR #209, opened 2026-04-13
Fix DIRKTimeStepper import

pbrubeck:


rckirby (approved):
Is this PR to encourage users to use the `stage_type` kwarg to `TimeStepper`, or is there something else more subtle here?

pbrubeck:
> Is this PR to encourage users to use the `stage_type` kwarg to `TimeStepper`, or is there something else more subtle here?

This item appeared twice in `__all__`. We were adding it before importing it and then we added it again.

Is this class not supposed to be public?

ScottMacLachlan:
> > Is this PR to encourage users to use the `stage_type` kwarg to `TimeStepper`, or is there something else more subtle here?
> 
> This item appeared twice in `__all__`. We were adding it before importing it and then we added it again.
> 
> Is this class not supposed to be public?

I stumbled on this trying to sort out a merge conflict when I started #202.  Just some book-keeping that needed to happen, I think.

That said, we worked in several spots to put all of these behind the `TimeStepper` interface, so it isn't clear to me that we're doing anybody any favours by exposing these.

rckirby:
> > This item appeared twice in `__all__`. We were adding it before importing it and then we added it again.
> > Is this class not supposed to be public?
> 
> I stumbled on this trying to sort out a merge conflict when I started #202. Just some book-keeping that needed to happen, I think.
> 
> That said, we worked in several spots to put all of these behind the `TimeStepper` interface, so it isn't clear to me that we're doing anybody any favours by exposing these.

It's fine with me for the particular stepper classes to be public, but I think it's better practice for users to use `TimeStepper` in case they want to switch/loop over various methods.  I don't know how pedantic/paternalistic we need to be here, however.  We have other steppers that aren't ready to incorporate behind `TimeStepper` (Nystrom is a current example, we will hopefully soon have exponential integrators).  We might want those in `__all__`.

Since there was a duplicate in `__all__`, I'm glad to just remove that since it's obviously wrong and we can entertain actual policy changes through an issue/discussion if desired.

PR #211, opened 2026-04-13
Test IrksomeImportOrderException

pbrubeck:


pbrubeck:
The test passes (failing to import irksome) on firedrake `main`, but not on `release`

PR #213, opened 2026-04-14
LinearVariationalProblem/Solver interface

pbrubeck:
Enables `TimeStepper(F, ...)` to take in bilinear forms `F = a(t; w, v) - L(t; v)` and to internally use a `LinearVariationalProblem/Solver`. Also adds the kwarg `constant_jacobian=True`, as an alternative to `snes_lag_jacobian -2` solver option.

### DIFF
--- a/demos/heat/demo_heat.py.rst
+++ b/demos/heat/demo_heat.py.rst
@@ -32,7 +32,7 @@ As usual, we need to import firedrake::
 
 We will also need to import certain items from irksome::
 
-  from irksome import GaussLegendre, Dt, MeshConstant, TimeStepper
+  from irksome import GaussLegendre, Dt, TimeStepper
 
 We will create the Butcher tableau for the lowest-order Gauss-Legendre
 Runge-Kutta method, which is more commonly known as the implicit
@@ -55,9 +55,8 @@ standard Firedrake fashion::
 
 We define variables to store the time step and current time value::
 
-  MC = MeshConstant(msh)
-  dt = MC.Constant(10.0 / N)
-  t = MC.Constant(0.0)
+  dt = Constant(10.0 / N)
+  t = Constant(0.0)
 
 This defines the right-hand side using the method of manufactured solutions::
 
@@ -84,10 +83,7 @@ standard UFL notation, augmented by the ``Dt`` operator from Irksome::
 
 Later demos will show how to use Firedrake's sophisticated interface
 to PETSc for efficient block solvers, but for now, we will solve the
-system with a direct method (Note: the matrix type needs to be
-explicitly set to aij if sparse direct factorization were used with a
-multi-stage method, as we wind up with a mixed problem that Firedrake
-will assemble into a PETSc MatNest otherwise)::
+system with a direct method. ::
 
   luparams = {"mat_type": "aij",
               "ksp_type": "preonly",
@@ -96,7 +92,7 @@ will assemble into a PETSc MatNest otherwise)::
 Most of Irksome's magic happens in the :class:`.TimeStepper`.  It
 transforms our semidiscrete form `F` into a fully discrete form for
 the stage unknowns and sets up a variational problem to solve for the
-stages at each time step.::
+stages at each time step. ::
 
   stepper = TimeStepper(F, butcher_tableau, t, dt, u, bcs=bc,
                         solver_parameters=luparams)
@@ -107,12 +103,48 @@ problem to compute the Runge-Kutta stage values and then updates the solution.::
 
   while (float(t) < 1.0):
       if (float(t) + float(dt) > 1.0):
-          dt.assign(1.0 - float(t))
+          dt.assign(1.0 - t)
       stepper.advance()
       print(float(t))
-      t.assign(float(t) + float(dt))
+      t.assign(t + dt)
 
 Finally, we print out the relative :math:`L^2` error::
+
+  print()
+  print(norm(u-uexact)/norm(uexact))
+
+So far we have neglected the fact that the problem is linear, and that the
+Jacobian is time-independent. We can instead formulate the semidiscrete form
+`F` in terms of a :class:`TrialFunction`, and specify the option
+``constant_jacobian``. So that the factorization is not redone at each time
+step. ::
+
+  dt.assign(10.0 / N)
+  t.assign(0)
+
+  u = Function(V)
+  u.interpolate(uexact)
+
+  v = TestFunction(V)
+  w = TrialFunction(V)
+  F = inner(Dt(w), v)*dx + inner(grad(w), grad(v))*dx - inner(rhs, v)*dx
+
+  linear_stepper = TimeStepper(F, butcher_tableau, t, dt, u, bcs=bc,
+                               constant_jacobian=True,
+                               solver_parameters=luparams)
+
+We can flag the Jacobian for an update by calling :meth:`invalidate_jacobian`,
+in case ``dt`` is changed ::
+
+  while (float(t) < 1.0):
+      if (float(t) + float(dt) > 1.0):
+          linear_stepper.invalidate_jacobian()
+          dt.assign(1.0 - t)
+      linear_stepper.advance()
+      print(float(t))
+      t.assign(t + dt)
+
+Finally, we see that we obtain same relative :math:`L^2` error from before::
 
   print()
   print(norm(u-uexact)/norm(uexact))
--- a/demos/heat/demo_heat.py.rst
+++ b/demos/heat/demo_heat.py.rst
@@ -116,7 +116,7 @@ Finally, we print out the relative :math:`L^2` error::
 So far we have neglected the fact that the problem is linear, and that the
 Jacobian is time-independent. We can instead formulate the semidiscrete form
 `F` in terms of a :class:`TrialFunction`, and specify the option
-``constant_jacobian``. So that the factorization is not redone at each time
+``constant_jacobian``, so that the factorization is not redone at each time
 step. ::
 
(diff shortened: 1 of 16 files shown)


demos/heat/demo_heat.py.rst
```diff
@@ -84,10 +83,7 @@ standard UFL notation, augmented by the ``Dt`` operator from Irksome::
 
 Later demos will show how to use Firedrake's sophisticated interface
 to PETSc for efficient block solvers, but for now, we will solve the
-system with a direct method (Note: the matrix type needs to be
-explicitly set to aij if sparse direct factorization were used with a
-multi-stage method, as we wind up with a mixed problem that Firedrake
-will assemble into a PETSc MatNest otherwise)::
```
pbrubeck, line 90: This isn't true. You only get a MatNest by default if one of the fields of the unknown is in the Real space. For these cases there is no way to get MatAIJ.

irksome/base_time_stepper.py
```diff
@@ -45,41 +50,43 @@ class StageCoupledTimeStepper(BaseTimeStepper):
...
-         `u0`.  This serves as a variable referring to the current time.
-    :arg dt: a :class:`Function` on the Real space over the same mesh as
-         `u0`.  This serves as a variable referring to the current time step.
-         The user may adjust this value between time steps.
+        ``F(t, u; v) == 0``, where ``u`` is the unknown
+        :class:`firedrake.Function` and ``v`` is the
+        :class:`firedrake.TestFunction`. To specify a linear problem,
+        ``F`` must be of the form ``a(t; w, v) - L(t; v)``, where
+        ``w`` is a :class:`firedrake.TrialFunction`.
+    :arg Fexp: A :class:`ufl.Form` instance describing the part of the
```
ScottMacLachlan: Looks like we lost the rest of this descriptor?

irksome/labeling.py
```diff
@@ -162,3 +163,20 @@ def as_form(form):
     if isinstance(form, LabelledForm):
         form = Form([]) if len(form) == 0 else form.form
     return form
+
+
+def as_linear_form(F, u0):
+    """
+    If `F` is a bilinear :class:`Form` compute a linear
+    :class:`Form` by replacing the trial function with `u0`.
```
ScottMacLachlan: Is it worth making clear (in either or both of the function name and this doctoring) that `as_linear_form` does nothing if `F` is not bilinear?

irksome/dirk_stepper.py
ScottMacLachlan, line 146: Why does `StageCoupledTimeStepper` have logic for whether this should be a `LinearVariationalProblem` or not, but DIRK not have it?

irksome/imex.py
```diff
@@ -205,9 +218,14 @@ def __init__(self, F, Fexp, butcher_tableau,
             Fexp, butcher_tableau, u0, UU_old, t, dt, splitting)
 
         self.itprob = NonlinearVariationalProblem(
```
ScottMacLachlan, line 220: Again, I'm confused about why we're not plumbing in the same `Linear` vs. `Nonlinear` logic here.

irksome/imex.py
```diff
@@ -414,12 +444,15 @@ def __init__(self, F, F_explicit, butcher_tableau, t, dt, u0, bcs=None,
         else:
             appctx = {**appctx, **appctx_irksome}
 
-        self.problem = NonlinearVariationalProblem(stage_F, k, bcnew)
+        self.problem = NonlinearVariationalProblem(stage_F, k, bcnew,
```
ScottMacLachlan, line 447: Here, again, seems inconsistent

irksome/imex.py
ScottMacLachlan, line 359: Since `replhat` depends on `dt`, I think we need to invalidate `mass_solver` when `dt` changes (i.e., when we invalidate `solver`

irksome/imex.py
```diff
@@ -596,3 +629,9 @@ def _finalize_stiffly_accurate(self):
 
     def solver_stats(self):
         return self.num_steps, self.num_nonlinear_iterations, self.num_linear_iterations, self.num_mass_nonlinear_iterations, self.num_mass_linear_iterations
+
+    def invalidate_jacobian(self):
+        """
+        Forces the matrix to be reassembled next time it is required.
+        """
+        LinearVariationalSolver.invalidate_jacobian(self.solver)
```
ScottMacLachlan, line 637:
```suggestion
        LinearVariationalSolver.invalidate_jacobian(self.solver)
        LinearVariationalSolver.invalidate_jacobian(self.mass_solver)
```

irksome/stage_derivative.py
```diff
@@ -18,35 +18,36 @@ def getForm(F, butch, t, dt, u0, stages, bcs=None, bc_type=None, splitting=AI, a
...
+        equation, or "ODE", which takes the time derivative of the
+        boundary data and evaluates this for the stage values.
+        Support for `firedrake.EquationBC` in `bcs` is limited
+        to DAE style BCs.
     :kwarg splitting: a callable that maps the (floating point) Butcher matrix
-         a to a pair of matrices `A1, A2` such that `butch.A = A1 A2`.  This is used
-         to vary between the classical RK formulation and Butcher's reformulation
-         that leads to a denser mass matrix with block-diagonal stiffness.
-         Some choices of function will assume that `butch.A` is invertible.
+        a to a pair of matrices `A1, A2` such that `butch.A = A1 A2`.  This is used
```
ScottMacLachlan:
```suggestion
        to a pair of matrices `A1, A2` such that `butch.A = A1 A2`.  This is used
```

irksome/stage_derivative.py
```diff
@@ -18,35 +18,36 @@ def getForm(F, butch, t, dt, u0, stages, bcs=None, bc_type=None, splitting=AI, a
...
+        to vary between the classical RK formulation and Butcher's reformulation
+        that leads to a denser mass matrix with block-diagonal stiffness.
+        Some choices of function will assume that `butch.A` is invertible.
     :kwarg aux_indices: a list of field indices to be discretized as :class:`TimeDerivative`,
-         analogouos to :class:`ContinouosPetrovGalerkinTimeStepper`.
+        analogouos to :class:`ContinouosPetrovGalerkinTimeStepper`.
```
ScottMacLachlan:
```suggestion
        analogous to :class:`ContinouosPetrovGalerkinTimeStepper`.
```

ScottMacLachlan (dismissed):
I'm a little confused about the completeness of this - it seems like we make an effort to use `LinearVariationalSolver` in `StageCoupledTimeStepper`, but nowhere else.  Is there a reason for that?

irksome/dirk_stepper.py
pbrubeck: LVS internally creates a NLVS but with different default solver options. The current code for DIRK will accept TrialFunction, but use a Newton SNES (and converge in one iteration). We can add it for DIRK,  but I see very little value.

irksome/imex.py
pbrubeck: But does `Fexp` depend on `khat`?

demos/heat/demo_heat.py.rst
```diff
@@ -107,12 +103,48 @@ problem to compute the Runge-Kutta stage values and then updates the solution.::
...
 
 Finally, we print out the relative :math:`L^2` error::
 
   print()
   print(norm(u-uexact)/norm(uexact))
+
+So far we have neglected the fact that the problem is linear, and that the
+Jacobian is time-independent. We can instead formulate the semidiscrete form
+`F` in terms of a :class:`TrialFunction`, and specify the option
+``constant_jacobian``. So that the factorization is not redone at each time
```
pbrubeck:
```suggestion
``constant_jacobian``, so that the factorization is not redone at each time
```

irksome/imex.py
ScottMacLachlan: No - it doesn't.

irksome/dirk_stepper.py
ScottMacLachlan: If there's no value for adding it for DIRK (and in other spots), then what's the value in adding it for `StageCoupledTimeStepper`?  Like I said above - I'm confused as to the goal here...

pbrubeck: These different steppers do not inherit from the same base class. I would argue that they should. Adding the LVS/NLVS branching all over the codebase didn't seem like the most easy to maintain solution. If we want consistency we should wrap this logic in a helper function

ScottMacLachlan: Right.  As currently structured, we have 3 separate DIRK-esque implementations, one for standard DIRK (includes the explicit case), one for DIRK-IMEX and one for Nystrom-DIRK (again including the explicit case).  So, this seems to expose that we should try to unify some structure there (although IMEX is probably a bit nastier to unify).

I guess that I'm okay with deferring that cleanup (and associated optimization) until after that's done, unless we think it's a smaller job than it seems at first glance.

pbrubeck: That sounds like the way to move forward. We should unify DIRKs and then it'd be easier to add the LVS optimization

irksome/base_time_stepper.py
pbrubeck: On the contrary, Fexp got copied over by mistake
```suggestion
```

irksome/imex.py
rckirby: We'll be handling this separately, per our discussion.

rckirby (approved):
I think this is step in the right direction for "typical" use cases to support `TrialFunction` inputs and get `LVP` underneath in Firedrake.  We'll need to think about some major refactorization before we can put this into per-stage methods like DIRK.

PR #214, opened 2026-04-14
update .gitignore

jdsteph2:
ignore built documentation files:

/docs/source/irksome.ufl.rst
/docs/source/irksome.backends.rst

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,8 @@
 /docs/build/
 /docs/source/demos/
 /docs/source/irksome.rst
+/docs/source/irksome.ufl.rst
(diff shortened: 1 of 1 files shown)


pbrubeck:
Shouldn't these files live in `docs/build`?

jdsteph2:
> Shouldn't these files live in `docs/build`?

Not sure.  We already have to ignore `/docs/source/irksome.rst`.

These started being tracked once I merged master into the multistep branch (and built the docs).

rckirby (approved):
My best guess is these popped up with auto-generation when we refactored to go toward a fenics backend and we can safely ignore them?

PR #222, opened 2026-05-05
DiscontinuousGalerkinCollocationScheme

pbrubeck:


irksome/scheme.py
```diff
@@ -126,6 +128,35 @@ def __init__(self, order,
...
+    def __init__(self, order,
+                 deriv_type="weak",
+                 quadrature_degree=None,
+                 quadrature_scheme="radau",
+                 max_quadrature_degree=None):
+        if order < 0:
+            raise ValueError(f"{type(self).__name__} must have order >= 0")
+        if quadrature_scheme is None:
+            basis_type = "spectral"
+        elif quadrature_scheme in {"radau", "lobatto"}:
```
pbrubeck, line 140: Should we only accept Radau, or do the other quadrature schemes still make sense?

irksome/scheme.py
```diff
@@ -126,6 +128,35 @@ def __init__(self, order,
                          max_quadrature_degree=max_quadrature_degree)
 
 
+class DiscontinuousGalerkinCollocationScheme(DiscontinuousGalerkinScheme):
+    """Class for describing collocation DG-in-time methods"""
+    def __init__(self, order,
+                 deriv_type="weak",
```
pbrubeck: The default should be consistent
```suggestion
                 deriv_type="strong",
```

irksome/discontinuous_galerkin_stepper.py
```diff
@@ -247,18 +250,37 @@ def __init__(self, F, scheme, t, dt, u0, bcs=None, basis_type=None,
         self.quadrature = quadrature
         self.max_quadrature_degree = scheme.max_quadrature_degree
 
-        num_stages = order+1
-
         self.update_b = vecconst(self.el.tabulate(0, (1.0,))[(0,)])
 
-        super().__init__(F, t, dt, u0, num_stages, bcs=bcs, **kwargs)
+        if isinstance(scheme, DiscontinuousGalerkinCollocationScheme):
```
pbrubeck: We should be checking here for Lagrange-type elements

irksome/discontinuous_galerkin_stepper.py
pbrubeck, line 21: Maybe we want Radau to be the default. And potentially make `DiscontinousGalerkinCollocationScheme` obsolete.

irksome/discontinuous_galerkin_stepper.py
```diff
@@ -247,18 +250,37 @@ def __init__(self, F, scheme, t, dt, u0, bcs=None, basis_type=None,
...
+        if tableau is not None:
+            # Galerkin collocation is equivalent to an IRK up to row scaling
+            row_scale = tableau.b
+
+            def scaledIA(A):
+                # For stage-value the splitting exposes row scaling
+                A1, A2 = IA(A)
+                np.multiply(A1, row_scale, out=A1)
+                np.multiply(1/row_scale, A2, out=A2)
+                return A1, A2
```
pbrubeck, line 280: For mismatching bases between the DG element and the tableau we might be able to introduce a basis transformation in the splitting.

irksome/scheme.py
rckirby: My guess is that we just want to do Radau for now.  That's the one I see equivalences with DG in the literature.

irksome/discontinuous_galerkin_stepper.py
rckirby: Lagrange-Radau seems to be the best default for DG.  It's "most like" RadauIIA methods.  Unlike cPG, which is a bit of a chameleon and can reproduce both Gauss & Radau methods with appropriate basis/quadrature choices.  

What does `DGCollocationScheme` being obsolete mean?  You've just introduced it on this PR?

irksome/discontinuous_galerkin_stepper.py
rckirby: Not sure we want that (yet).  For now, it's more important for there to be A Way to get an equivalent IRK scheme than for us to be able to do this for every DG formulation.  There are costs associated with the basis transformation, and we might wait till there is an actual use case.

irksome/scheme.py
rckirby: Now may be to revisit bikeshedding on names.  Both cPG collocation and DG collocation are being used to refer to particular basis choices that could lead to collocation schemes, although they are only properly collocated if one selects the correct quadrature rule.  What's more, we've got very different basis flavors for CPG (integrated Lagrange trial/ Lagrange test) versus DG (Lagrange) that lead to quite different IRK formulations (stage derivative vs value, etc).  The idea of a convenience class that bakes in the "right" basis choices to get such an equivalence (but allows quadrature flexibility) seems like a great idea, I'm just not sure our current naming scheme expresses what we want.

tests/test_pc.py
```diff
@@ -7,24 +7,27 @@
...
-def Fubc(V, t, uexact, Lhigh=None):
-    if Lhigh is None:
-        Lhigh = lambda x: x
+def Fubc(V, t, uexact):
     u = Function(V)
     u.interpolate(uexact)
     v = TestFunction(V)
     rhs = Dt(uexact) - div(grad(uexact)) - uexact * (1-uexact)
-    F = inner(Dt(u), v)*dx + inner(grad(u), grad(v))*dx + Lhigh(-inner(u*(1-u), v)*dx - inner(rhs, v)*dx)
+    F = inner(Dt(u), v)*dx + inner(grad(u), grad(v))*dx - inner(u*(1-u), v)*dx - inner(rhs, v)*dx
```
rckirby, line 30: labels going because we can do `"auto"` for quadrature and compare to a fixed low degree?

rckirby (commented):
Thanks for getting this started.  See comments/questions

irksome/discontinuous_galerkin_stepper.py
pbrubeck: Sure, this can be part of a later PR. I just wanted raise awareness.

irksome/discontinuous_galerkin_stepper.py
pbrubeck: Currently, DGCollocationScheme is almost identitcal to DGScheme with basis_type="radau". The only difference is in how the default quadrature degree is set. So we should consider whether we want to add a whole new class just to change this default.

irksome/scheme.py
pbrubeck: We should keep the convenience class. It'd be hard for users to select all these options themselves. But I would advocate for making these convenience classes the default.

tests/test_pc.py
pbrubeck: Labels can go because we are now setting `quad_degree="auto"`. The preconditioner for these tests is always an IRK, which entirely bypasses the underintegrated Galerkin scheme

irksome/discontinuous_galerkin_stepper.py
rckirby: Ah, the argument for keeping a (very) light class is a parallel naming scheme with cPG?

irksome/scheme.py
rckirby: We're agreed on the convenience class, and making it the default seems natural.  But I'm objecting to the fact that if you don't use exactly the right quadrature rule, it's not a collocation scheme, despite that being in the name of the class.

tests/test_pc.py
rckirby: Very good, thanks for confirming.

tests/test_pc.py
```diff
@@ -192,15 +204,20 @@ def test_git_irk_equivalence(quad_scheme, order, stage_type):
...
+            "mat_type": "nest",
+            "sub_mat_type": "aij",
+            "pc_type": "cholesky" if scheme.num_stages == 1 else "lu",
             "pc_factor_mat_solver_type": "mumps",
         }
     }
-    scheme = GalerkinCollocationScheme(order, quadrature_scheme=quad_scheme, stage_type=stage_type)
+
+    kwargs = {}
+    if isinstance(scheme, GalerkinCollocationScheme):
```
pbrubeck:
```suggestion
    if isinstance(scheme, ContinuousPetrovGalerkinScheme):
```

irksome/discontinuous_galerkin_stepper.py
pbrubeck: Agree

rckirby (approved):

PR #223, opened 2026-05-06
Release workflow

pbrubeck:


.github/workflows/pypi.yml
```diff
@@ -0,0 +1,67 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: Git ref to publish
+        type: string
+      pypi:
+        description: Publish to PyPI
+        default: false
```
connorjward: Not sure where you got these workflows from originally but I would recommend not having a separate PyPI workflow. Just do everything in `release.yml`. The workflow from [petsctools](https://github.com/firedrakeproject/petsctools/blob/main/.github/workflows/release.yml) is basically what you need.

Also I'd recommend not using `secrets` and instead use [trusted publishing](https://docs.pypi.org/trusted-publishers/). In general it's the better thing to do.

connorjward (dismissed):

.github/workflows/pypi.yml
pbrubeck: I just copied these from FIAT

pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "IRKsome"
-version = "0.0.1"
+version = "0.1.0"
```
pbrubeck: @connorjward is this okay or should we go for something more meaningful? Could we include just the year as in petsctools?

.github/workflows/pypi.yml
connorjward: I have written about 8 of these now. The FIAT one is slightly not my recommended approach any more.

pyproject.toml
connorjward: I think a date based version number is a better idea. Since you will want to have main and release branches you will need to have both major and minor versions (in case you have to patch a previous release). I'd recommend having major versions like
```
2026.0.0
2026.1.0
...
```
and minor versions like
```
2026.0.0
2026.0.1
2026.0.2
...
```

You could also go for the month in the second slot. I.e. 2026.4.0 instead of 2026.0.0. That's just a taste thing. Only thing is to avoid doing `2026.04` because leading zeros are ignored by PyPI.

pbrubeck:
```suggestion
version = "2026.0.0"
```

pbrubeck: Thanks. We discussed and for now will keep a single branch, so it is easier to maintain. So far, even though Firedrake has gone through significant API-breaking changes, these haven't caused the actual need to split into two Irksome branches. We always had Irksome working on either Firedrake branch.

connorjward: I still recommend this versioning strategy because then you'll be prepared in case you ever have to.

pbrubeck: Sure, I applied your suggestion

PR #224, opened 2026-05-06
Add kronpc for stage-coupled Kronecker product preconditioners

436ahsan:
## Summary

This PR adds a `kronpc` framework for applying stage-coupled Kronecker product preconditioners of the form

\[
L \otimes K^{-1},
\]

where \(L\) is a small dense stage matrix, typically derived from the Runge--Kutta Butcher matrix, and \(K^{-1}\) is supplied by a single-stage PETSc/Firedrake subsolver.

The main motivation is to support Runge-Kutta stage preconditioners appearing in augmented Lagrangian Schur-complement approximations for time-dependent Navier-Stokes problems.

This allows users to build preconditioners involving Kronecker products of Runge-Kutta stage matrices with pressure, mass, and stiffness inverses. This structure is motivated by the augmented Lagrangian Runge-Kutta preconditioner of Leveque, He, and Olshanskii, arXiv:2506.04451.

## Changes

- Adds `irksome/kronpc.py`.
- Provides:
  - `KronPC`
  - `MassKronPC`
  - `StiffnessKronPC`
  - `SIPGStiffnessKronPC`
- Re-exports these classes from `irksome.__init__`.
- Adds a test for `MassKronPC` against a full monolithic LU solve.
- Updates the implementation for the current petsc4py `KSP.setDMActive` API.

## Testing

```bash
OMP_NUM_THREADS=1 python -m pytest tests/test_kronpc.py

### CODE REVIEW

File: irksome/kronpc.py [hunk 1]
```diff
@@ -0,0 +1,321 @@
...
+        # Stagewise K^{-1}
+        for i in range(s):
+            with self.work_in.subfunctions[i].dat.vec_ro as rhs_i, \
+                    self.work_mid.subfunctions[i].dat.vec_wo as mid_i:
+                mid_i.set(0.0)
+                self.sub_ksp.solve(rhs_i, mid_i)
+
+        # Zero outputs
+        for j in range(s):
+            self.work_out.subfunctions[j].assign(0.0)
```
pbrubeck:
```suggestion
        self.work_out.zero()
```

436ahsan: You are right @pbrubeck , this is cleaner because it zeros the whole mixed/stage-stacked function at once instead of looping through subfunctions.

tests/test_kronpc.py
```diff
@@ -0,0 +1,85 @@
...
+    rng = np.random.default_rng(seed)
+    L = Cofunction(Vbig.dual(), name="rhs")
+    with L.dat.vec_wo as v:
+        arr = rng.standard_normal(v.getSize())
+        v.setValues(range(v.getSize()), arr)
+        v.assemble()
+    return L
+
+
+@pytest.mark.parametrize("ns", [2, 3, 4, 5])  # meaningful cases are ns >= 2
```
rckirby: Going up to 5 stages in a test may be overkill for testing correctness and adding to run-time on the test suite?

tests/test_kronpc.py
```diff
@@ -0,0 +1,85 @@
...
+def test_mass_kron_pc(ns):
```
rckirby: Is it possible to get this to test the stiffness PCs as well with a parameterization?  You'd need to parametrize over the PC paired with a form that did mass vs stiffness.

rckirby (changes requested):
Thanks for the feature.  We might want to add a demo of the Leveque et al preconditioner using this for flow later, but for now we probably want to just make sure both mass & stiffness get tested.

tests/test_kronpc.py
436ahsan: Yes, that makes sense. I reduced the test to ns = [2, 3], which still checks nontrivial stage coupling while keeping the test cheaper for CI.

tests/test_kronpc.py
436ahsan: I parameterized the KronPC test over both MassKronPC and StiffnessKronPC. The stiffness case is paired with the corresponding shifted stiffness form A \otimes (K + shift M). I use a larger shift only in the test to keep the reference operator well-conditioned; the default remains 1e-12. I also tried adding SIPGStiffnessKronPC, but the current SIPG implementation projects out a constant nullspace in apply(), while the SIPG form includes boundary penalty terms. I think for this reason there are still some issues regarding passing the test with SIPG, and I need to think more about how to resolve this issue.

PR #225, opened 2026-05-07
Support for multidomain problems

pbrubeck:
Suppose `u = (u0, u1) in V0 * V1` is the unknown for a mixed formulation where the function spaces `V0` and `V1` are defined on different meshes. At stage `i` we replace `u -> (A*k)[i]`, where `k[i] = (k[0][i], k[1][i])` is the component of the stage vector for stage i.

For a single-domain integral like `inner(u[0], v[0]) * dx(mesh0)`, we run into an issue when we replace `u[0] -> (A*k)[i][0]` inside the integral as `k` is defined on both `mesh0` and `mesh1`.

This PR fixes this by calling `expand_indices` which eventually does the subsitution `u[0] -> (A*k[0])[i]`, so `k[1]` no longer appears inside the `dx(mesh0)` integral.

This seems more like a Firedrake bug, so maybe we should be calling `expand_indices` early enough in the form compiler.

### DIFF
--- a/tests/test_multidomain.py
+++ b/tests/test_multidomain.py
@@ -1,0 +1,37 @@
+from firedrake import *
+from irksome import GaussLegendre, Dt
+from irksome.stage_derivative import getForm
+import pytest
+
+
+@pytest.mark.parametrize("num_stages", [1, 2])
+def test_multidomain_getForm(num_stages):
+    t = Constant(0)
+    dt = Constant(0.1)
+    mesh1 = UnitSquareMesh(1, 1)
+    mesh2 = Submesh(mesh1, mesh1.topological_dimension-1, 1, label_name="exterior_facets")
+
+    V1 = FunctionSpace(mesh1, "CG", 1)
+    V2 = FunctionSpace(mesh2, "CG", 1)
+    Z = V1 * V2
+
+    z = Function(Z)
+    u1, u2 = split(z)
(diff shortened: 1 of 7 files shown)


tests/test_multidomain.py
```diff
@@ -0,0 +1,37 @@
+from firedrake import *
+from irksome import GaussLegendre, Dt
+from irksome.stage_derivative import getForm
+import pytest
+
+
+@pytest.mark.parametrize("num_stages", [1, 2])
+def test_multidomain_getForm(num_stages):
```
pbrubeck, line 8: We should add a nontrivial MMS test

tests/test_multidomain.py
```diff
@@ -0,0 +1,37 @@
...
+    v1, v2 = TrialFunctions(Z)
+
+    dx1 = dx(mesh1)
+    dx2 = dx(mesh2)
+    dx12 = Measure("ds", domain=mesh1, intersect_measures=[dx2])
+
+    F = (inner(Dt(u1), v1) * dx1
+         + inner(Dt(u2), v2) * dx2
+         - inner(u1, v1) * dx1
+         - inner(u2, v2) * dx2
```
pbrubeck:
```suggestion
         + inner(u1, v1) * dx1
         + inner(u2, v2) * dx2
```

PR #226, opened 2026-05-13
Mass-conservative update for non-stiffly-accurate stage_value

sghelichkhani:
## Summary

@rckirby — this is the follow-up to our Slack conversation about implicit midpoint and mass conservation. Your hand-derivation was right and there was a bug, just subtler than "the discretisation is wrong". The per-stage equations are fine; the bug lives in the update step that produces `u_new` from the stages.

## Diagnosis

For `inner(Dt(g(u)), v)*dx + spatial(u)` with nonlinear `g` (Richards moisture content `theta(h)` is the canonical case), PR #204 already builds the per-stage equations as the two-evaluation form `g(U_i) - g(u_0)` inside `getFormStage`. So the stage-level conservation property is correct for any RK method — exactly what your hand-derivation predicts at the PDE level.

Conservation was still being lost for non-stiffly-accurate methods (`GaussLegendre(1)` = implicit midpoint, `GaussLegendre(2)`, `QinZhang`). The cause is in `_update_Ainv`: after the stages are computed, Irksome forms `u_new = (1 - sum bAinv) u_0 + sum bAinv U_i`, a linear combination of `u_0` and the stage values. For linear `Dt(u)` that is algebraically equivalent to the conservative time integration. For `Dt(theta(h))` it is not: `theta` of a linear combination is not the linear combination of per-stage `theta` increments, so the conservation property the stage equations establish is destroyed at the update step. Stiffly accurate methods avoid this entirely because `u_new = U_s` is identically the last stage value, so the conservative finite difference sits in that stage's equation and gets copied straight to `u_new`. That is why RadauIIA, LobattoIIIA/C, BackwardEuler and Alexander all look fine and only the non-SA methods flagged.

## Fix

Route the non-SA path through a conservative variational solve when the form has a composite `Dt`:

```
replace(F_dtless, {u_0: u_new}) - replace(F_dtless, {u_0: u_0})
    + dt * sum_i b_i * F_rem(stage_i) = 0
```

solved for `u_new`. For `g = identity` it collapses to the original `inner(u_new - u_0, v)*dx` head, so nothing changes for the linear case. For nonlinear `g` it is a discrete statement of the global conservation invariant and `u_new` satisfies it by construction. Order is preserved: the stages are still order `p`, the RK quadrature on the spatial term is unchanged, and inverting the conservation equation for `u_new` preserves order provided `g` is sufficiently smooth.

Linear `Dt(u)` deliberately keeps `_update_Ainv`. The new path is algebraically identical to the linear-combination update for `g = identity` (verified by direct calculation), and `_update_Ainv` handles DAE structure correctly while the conservative variational head does not (the algebraic block is unconstrained and the linear solve is singular). Composite-Dt detection is the right syntactic gate for routing — it lives in `irksome/ufl/manipulation.py` as `has_composite_time_derivative`, with documented exemptions for structural views, linear differential operators, DG facet operators, and arithmetic that is linear in `u_0`.

## What is intentionally not in this PR

In line with your earlier comment about DIRKs — that we run DIRK in stage-derivative form today but could plausibly do a stage-value form for them at some point, and that the right play is to push fully-implicit solvers in `stage_value` and let accuracy buy back the runtime — this PR does not touch `DIRKTimeStepper`. Stage-derivative DIRK folds the chain rule into the K-substitution and has no clean conservative analogue without restructuring; users who need conservation can run any DIRK tableau through `stage_type='value'` instead. There is a telescoping W-ladder fix on a separate branch but it belongs to its own conversation.

## Test

The existing `tests/test_mass_conservation.py` already runs the exponential Richards problem with a flux BC and asserts mass error near machine precision. I extended the parametrize list to include `GaussLegendre(1)`, `GaussLegendre(2)` and `QinZhang` alongside the existing BackwardEuler / RadauIIA(2) / BDF / AdamsMoulton entries.

Master Irksome with the three new entries:

| Scheme | Mass error |
|---|---|
| BackwardEuler, RadauIIA2, BDF1, AM0..2 | machine precision |
| **ImplicitMidpoint** | **8.88e-08** |
| **GaussLegendre(2)** | **4.78e-09** |
| **QinZhang** | **3.94e-09** |

This branch: all nine sit at machine precision.

The failure on master is the bug; the pass on this branch is the fix. The seven-tableau spread covers single-stage and multi-stage non-SA methods, order 2 and order 4, plus the negative-`b` Alexander case as a sign regression on the SA path.

`tests/test_has_composite_time_derivative.py` pins both directions of the classifier contract: linear-arithmetic forms (`Dt(c*u)`, `Dt(u/c)`, `Dt(u + sin(t))`) must not be flagged (else DAEs break), nonlinear forms (`Dt(u*u)`, `Dt(theta(u))`, `Dt(1/u)`) must be flagged (else non-SA `stage_value` silently mis-discretises).

Wider regression: `tests/test_dirk`, `test_dae`, `test_explicit`, `test_galerkin`, `test_disc_galerkin`, `test_split`, `test_time_form_splitting`, `test_pc`, `test_bern`, `test_butcher`, `test_differentiation`, `test_curl` all pass.

## Cross-validation outside Irksome

The same test runs through G-ADOPT's Richards solver wrapper on a Haverkamp soil model: a function-space sweep on BackwardEuler is the existing `test_mass_balance` test, and a new `test_mass_balance_tableaux` parametrising over BE, ImplicitMidpoint and GaussLegendre(2) at DQ1 produces the same master-fail / branch-pass pattern. File:

https://github.com/g-adopt/g-adopt/blob/48a8cf44/tests/richards/test_mass_balance.py#L145-L168


irksome/stage_value.py
```diff
@@ -220,20 +252,33 @@ def _update_stiff_acc(self):
...
+        unew = Function(u0.function_space())
+
         split_form = split_time_derivative_terms(F, t=t, timedep_coeffs=(u0,))
+        F_dtless = remove_time_derivatives(split_form.time)
         F_remainder = expand_time_derivatives(split_form.remainder, t=t, timedep_coeffs=())
+
+        # Two-evaluation conservative head.  Subtracting the two
+        # replaced forms keeps the test function v unchanged in both,
+        # so the result is a single Form valid for assembly.
+        Fupdate = replace(F_dtless, {u0: unew}) - replace(F_dtless, {u0: u0})
```
pbrubeck:
```suggestion
        Fupdate = replace(F_dtless, {u0: unew}) - F_dtless
```

irksome/stage_value.py
```diff
@@ -220,20 +252,33 @@ def _update_stiff_acc(self):
...
+        # Build a conservative variational update for u_new from the
+        # stage values.  The head is the same two-evaluation form used
+        # in the stage equations, evaluated at unew and u0:
+        #
+        #     replace(F_dtless, {u0: unew}) - replace(F_dtless, {u0: u0})
+        #
+        # which expands to inner(g(unew) - g(u0), v)*dx for the typical
+        # mass term inner(Dt(g(u)), v)*dx.  For g = identity this is
+        # exactly inner(unew - u0, v)*dx -- the previous head -- so the
+        # discrete update equation is unchanged in the linear case.
```
pbrubeck: Should this become part docstring and part comment?

irksome/stage_value.py
```diff
@@ -220,20 +252,33 @@ def _update_stiff_acc(self):
...
         u0 = self.u0
+        unew = Function(u0.function_space())
+
         split_form = split_time_derivative_terms(F, t=t, timedep_coeffs=(u0,))
+        F_dtless = remove_time_derivatives(split_form.time)
         F_remainder = expand_time_derivatives(split_form.remainder, t=t, timedep_coeffs=())
+
+        # Two-evaluation conservative head.  Subtracting the two
+        # replaced forms keeps the test function v unchanged in both,
+        # so the result is a single Form valid for assembly.
```
pbrubeck: Is this comment needed given the very detailed explanation above?

irksome/stage_value.py
```diff
@@ -258,6 +303,13 @@ def get_update_solver(self, update_solver_parameters):
         return unew, update_solver
 
     def _update_general(self):
+        # Seed unew with u0 before solving.  The conservative head's
+        # Jacobian is the moisture-capacity-weighted mass matrix
+        # C(unew)*v*phi*dx, and for soils like Haverkamp the capacity
+        # vanishes at h = 0 -- so a fresh Function (zero-initialised)
+        # gives a singular initial Jacobian and SNES fails before
+        # taking any Newton step.  u0 is the natural warm start anyway.
```
pbrubeck: Can we make this a docstring?

irksome/ufl/manipulation.py
```diff
@@ -11,19 +11,119 @@
...
+       resulting discretisation is *not* mass-conservative.  Always
+       wrap the symbolic expression directly in ``Dt`` (as
+       ``Dt(theta(u))``, not ``Dt(theta_function)``).
+    """
+    for td in extract_type(F, TimeDerivative):
+        f, = td.ufl_operands
+        if not _depends_on(f, u0):
+            # Dt(f(t,x)) -- chain-ruled analytically, no u0 dependence
+            continue
+        if _is_pass_through(f, u0):
```
pbrubeck: So here we would like to detect linear expressions in `u0`? I think UFL might already have similar utility functions to detect whether a `Form` is linear in its `Arguments`. I'll have a look and suggest a built in alternative

irksome/stage_value.py
```diff
@@ -184,25 +186,55 @@ def __init__(self, F, butcher_tableau, t, dt, u0, bcs=None,
                          sample_points=sample_points,
                          **kwargs)
 
+        # Conservative variational update inherits the stage solve's
+        # parameters by default: same form structure, same Jacobian
+        # sparsity, so the same preconditioner is the right starting
+        # point.  Callers who want a different setup can pass
+        # update_solver_parameters explicitly.
```
pbrubeck: Is the structure in the update solver always the same as the Jacobian, or is this only true for the conservative case?

irksome/stage_value.py
```diff
@@ -258,6 +303,13 @@ def get_update_solver(self, update_solver_parameters):
         return unew, update_solver
 
     def _update_general(self):
+        # Seed unew with u0 before solving.  The conservative head's
+        # Jacobian is the moisture-capacity-weighted mass matrix
+        # C(unew)*v*phi*dx, and for soils like Haverkamp the capacity
```
pbrubeck: We should remove the application-specific language from here

irksome/stage_value.py
pbrubeck: I just realized this whole paragraph is referring to `unew.assign(u0)`. I think this issue can be explained in fewer lines.

```suggestion
        # Constant-in-time initial guess to prevent singular Jacobian
```

irksome/ufl/manipulation.py
```diff
@@ -11,19 +11,119 @@
...
+                   for c, dep in zip(f.ufl_operands, deps))
+    if isinstance(f, Division):
+        num, den = f.ufl_operands
+        if _depends_on(den, u0):
+            return False
+        return (not _depends_on(num, u0)) or _is_pass_through(num, u0)
+    return False
+
+
+def has_composite_time_derivative(F, u0):
```
pbrubeck: Maybe nonlinear instead of composite?

irksome/stage_value.py
```diff
@@ -220,20 +252,33 @@ def _update_stiff_acc(self):
...
-        unew = Function(self.u0.function_space())
-        Fupdate = inner(unew - self.u0, v) * dx
-
-        C = vecconst(self.butcher_tableau.c)
-        B = vecconst(self.butcher_tableau.b)
+        # Build a conservative variational update for u_new from the
+        # stage values.  The head is the same two-evaluation form used
+        # in the stage equations, evaluated at unew and u0:
+        #
+        #     replace(F_dtless, {u0: unew}) - replace(F_dtless, {u0: u0})
```
pbrubeck: F_dtless is an internal concept, should we refer to `inner(Dt(g(u)), v)*dx` in the docstring and have here instead
```suggestion
        #     inner(g(unew) - g(u0), v) * dx
```

irksome/stage_value.py
sghelichkhani: Applied in fb28b88, thanks.

irksome/stage_value.py
sghelichkhani: Moved the derivation into the `get_update_solver` docstring in fb28b88.

irksome/stage_value.py
sghelichkhani: Dropped in fb28b88.

irksome/stage_value.py
sghelichkhani: Compressed to a single line in fb28b88, following your suggestion below.

irksome/ufl/manipulation.py
sghelichkhani: Had a look at `ufl/algorithms/check_arities.py`. Replicating this through `ArityChecker` might be more complicated than it sounds, unless I am missing something. The linear-operator coverage matches what our walker does, but its `sum` rule rejects `arg + constant` as an arity mismatch — so `Dt(u + sin(t))` and `Dt(2*u + 3)` fail, where for our purpose `Dt(u + c) = Dt(u) + Dt(c)` is fine. We would need to subclass to relax `sum` for the "one side has no arguments" case, at which point we are most of the way back to the walker. Verified on a small prototype: the multiplicative cases (`u*u`, `exp(u)`, `1/u`) work via a sentinel `Argument` substitution; the additive-with-constant case is the only divergence. Happy to swap if you see a clean way through.

irksome/stage_value.py
sghelichkhani: Good catch — the inheritance only matters for the conservative path. Moved it inside that branch in fb28b88 so it is scoped to where it is actually consumed.

irksome/stage_value.py
sghelichkhani: Removed in fb28b88.

irksome/stage_value.py
sghelichkhani: Applied verbatim in fb28b88, thanks.

irksome/ufl/manipulation.py
sghelichkhani: Renamed to `has_nonlinear_time_derivative` in fb28b88. Test file renamed in lockstep.

irksome/stage_value.py
sghelichkhani: Yes, agreed `F_dtless` is internal noise. Moved the user-facing algebraic form (`inner(g(unew) - g(u0), v) * dx`) into the `get_update_solver` docstring in fb28b88.

irksome/ufl/manipulation.py
pbrubeck: You are right, being linear is not the same as being affine. I think that we could make the sentinel `Argument` substitution, then apply `ufl.lhs`, and then check the arity.

pbrubeck: In the end we always want a linear problem for the update solver. Right now we create a nonlinear problem with `unew`. What if instead of defining the residual for the update we define the bilinear form with a `TrialFunction`? Then `LinearVariationalSolver` will automatically complain if we had nonlinearities in the `TrialFunction`

pbrubeck: But we need the check much before having to define the conservative solver.

irksome/stage_value.py
```diff
@@ -220,20 +238,33 @@ def _update_stiff_acc(self):
...
         t = self.t
         dt = self.dt
         u0 = self.u0
+        unew = Function(u0.function_space())
+
         split_form = split_time_derivative_terms(F, t=t, timedep_coeffs=(u0,))
+        F_dtless = remove_time_derivatives(split_form.time)
         F_remainder = expand_time_derivatives(split_form.remainder, t=t, timedep_coeffs=())
+
+        Fupdate = replace(F_dtless, {u0: unew}) - F_dtless
```
pbrubeck: Are we always expecting `F_dtless` to be linear in u0?

pbrubeck: No right?

irksome/ufl/manipulation.py
sghelichkhani: Right on the third point — the check is upfront, for dispatch. On `ufl.lhs`: it walks UFL nodes it knows about, and `Dt` is Irksome's own `TimeDerivative`, so `lhs` would only do something useful after `expand_time_derivatives` has already chain-ruled `Dt(g(u))` to `g'(u) * Dt(u)` — which is precisely the rewrite we are trying to gate on. Wrong order. On the LinearVariationalSolver route: the update head `g(u_new) - g(u_0)` is genuinely nonlinear in `u_new` for nonlinear `g`, so we need the nonlinear solve. So unless something else turns up, I would keep the walker.

irksome/stage_value.py
sghelichkhani: Correct — `F_dtless = inner(theta(u_0), v) * dx` in the canonical case, nonlinear in `u_0`. That is why the update solve is nonlinear.

irksome/stage_value.py
```diff
@@ -189,20 +190,37 @@ def __init__(self, F, butcher_tableau, t, dt, u0, bcs=None,
...
+            # linear case we keep the bAinv shortcut: it is conservative
+            # by construction (g = id makes both formulations agree)
+            # AND it handles DAEs correctly, which the conservative
+            # variational update does not.
+            if has_nonlinear_time_derivative(F, u0):
+                # The update solve uses the same form residual as the
+                # stage solve, posed on V rather than V^s.  Inherit
+                # solver parameters; callers can override.
+                if update_solver_parameters is None:
+                    update_solver_parameters = solver_parameters
```
pbrubeck: There's a potential issue here. What if the solver options for the stage problem use fieldsplit?

irksome/ufl/manipulation.py
pbrubeck: but couldn't we apply `ufl.lhs(ufl.replace(F_dtless, {u0: TrialFunction(V)}))`?

sghelichkhani: Fully agreed on the lhs-equality formulation. After substituting Trial for u0, `inner(Trial + sin(t), v) * dx` puts the `sin(t) * v` piece on `rhs`, so the `lhs == replaced form` test would flag `Dt(u + sin(t))` as nonlinear -- same affine-failure mode as the ArityChecker prototype.

But your underlying point is right: maintaining a parallel exemption list inside Irksome is a liability, and UFL already knows the linearity of `Grad`, `Div`, `Indexed`, `ListTensor`, `ComponentTensor`, restrictions, `CellAvg`, `FacetAvg`. The cleaner version of your idea is probably the Gateaux derivative test? differentiate `f` with respect to `u0` in a Trial direction, and check whether the derivative still depends on `u0`. If it does, `f` is nonlinear in `u0`.

```python
def has_nonlinear_time_derivative(F, u0):
    Trial = TrialFunction(u0.function_space())
    for td in extract_type(F, TimeDerivative):
        f, = td.ufl_operands
        if u0 not in extract_coefficients(f):
            continue
        D = expand_derivatives(derivative(f, u0, Trial))
        if u0 in extract_coefficients(D):
            return True
    return False
```

One UFL call, no exemption list. Passes the full `tests/test_has_nonlinear_time_derivative.py` matrix (`Dt(2u)`, `Dt(u/2)`, `Dt(u + sin(t))`, `Dt(2u + 3)` all linear; `Dt(u*u)`, `Dt(theta(u))`, `Dt(1/u)` all nonlinear), plus the structural-view, linear-differential-operator, and DG-facet-restriction cases the walker exempted explicitly. The full mass-conservation suite still passes at machine precision across all nine tableaux.

Landed in 68de837.

irksome/stage_value.py
sghelichkhani: I see! Agreed! Dropped the inheritance. When `update_solver_parameters` is None, the update solve now uses a fixed default tuned to its own structure rather than copying the stage parameters.

The default is CG + bjacobi/icc rather than LU:

```python
_DEFAULT_UPDATE_SOLVER_PARAMETERS = {
    'snes_type': 'newtonls',
    'ksp_type': 'cg',
    'pc_type': 'bjacobi',
    'sub_pc_type': 'icc',
}
```

Reasoning for this: the update Jacobian is `g'(u_new) * v * phi * dx` so the RK quadrature contributes spatial terms at known stage values, constant in `u_new`, so they drop from the linearisation. The whole operator is a weighted mass matrix: SPD when `g' > 0`, condition number bounded independent of mesh size, so CG converges in O(1) iterations with any cheap preconditioner. LU works at small sizes but does not scale; the iterative default is strictly better across the size range and loses nothing at the small end. `bjacobi + icc` picks up block structure for vector / mixed V automatically and degenerates gracefully on scalar V.

Two things I would value your view on:

1. Is there an Irksome-wide convention for default `solver_parameters`? I would rather match an existing pattern than introduce a one-off.
2. CG + bjacobi/icc bakes in monotone `g` (`g' > 0` a.e.). Non-monotone `g` breaks SPD and needs an explicit override (GMRES + a more general preconditioner). Comfortable with that assumption as the default, or would you prefer GMRES + bjacobi/ilu as the conservative choice at the cost of slightly slower convergence on the canonical case?

Docstring on `get_update_solver` now states the no-inheritance asymmetry explicitly so it does not surprise anyone.

pbrubeck: Irksome does not change the firedrake default solver parameters. I don't see why we should be changing them in any particular instance.

Firedrake by default uses a sparse direct solver, depending on which packages are installed. If the user wants an iterative solver they should set it explicitly.

irksome/ufl/manipulation.py
pbrubeck: Very neatly done, thank you

irksome/ufl/manipulation.py
```diff
@@ -11,19 +11,62 @@
...
+       under ``Dt`` after differentiation.  If a user creates an
+       intermediate :class:`~firedrake.Function` whose values were
+       interpolated from an expression in u0 and then writes
+       ``Dt(that_intermediate)``, the syntactic dependence on u0 is
+       lost and this function will declare the form safe.  The
+       resulting discretisation is *not* mass-conservative.  Always
+       wrap the symbolic expression directly in ``Dt`` (as
+       ``Dt(theta(u))``, not ``Dt(theta_function)``).
+    """
+    Trial = TrialFunction(u0.function_space())
```
pbrubeck:
```suggestion
```

irksome/ufl/manipulation.py
```diff
@@ -11,19 +11,62 @@
...
+    for td in extract_type(F, TimeDerivative):
+        f, = td.ufl_operands
+        if u0 not in extract_coefficients(f):
+            # Dt(f(t,x)) -- no u0 dependence, chain-ruled analytically
+            continue
+        D = expand_derivatives(derivative(f, u0, Trial))
```
pbrubeck:
```suggestion
        D = expand_derivatives(derivative(f, u0))
```

irksome/ufl/manipulation.py
```diff
@@ -11,19 +11,62 @@
 from operator import or_
 from typing import NamedTuple, Sequence, FrozenSet
 
+from ufl import TrialFunction, derivative
```
pbrubeck:
```suggestion
from ufl import derivative
```

irksome/stage_value.py
sghelichkhani: All sensible. Now reverted in 05e0d8e -- update_solver_parameters defaults to None, which lets NonlinearVariationalSolver pick up Firedrake's default. Test updated to pass tight tolerances through explicitly.

irksome/stage_value.py
```diff
@@ -189,20 +190,30 @@ def __init__(self, F, butcher_tableau, t, dt, u0, bcs=None,
...
-                b = butcher_tableau.b
-                self.bAinv = vecconst(numpy.linalg.solve(A.T, b))
-                self.update_scale = 1-numpy.sum(self.bAinv)
-                self._update = self._update_Ainv
-            except numpy.linalg.LinAlgError:
+            # For nonlinear Dt(g(u)) we need the conservative variational
+            # update; the bAinv linear combination of stages is correct
+            # for g = identity but breaks for nonlinear g.  For the
+            # linear case we keep the bAinv shortcut: it is conservative
+            # by construction (g = id makes both formulations agree)
```
pbrubeck: This comment makes me think that the linear case where `g = scalar * id` might still have some issues. Could you confirm?

irksome/stage_value.py
pbrubeck:
```suggestion
from ufl import Form, as_tensor, as_ufl
```

pbrubeck:
```suggestion
        update_problem = backend_cls.create_variational_problem(Fupdate, unew, update_bcs)
```

irksome/stage_value.py
```diff
@@ -220,22 +233,49 @@ def _update_stiff_acc(self):
...
 
         split_form = split_time_derivative_terms(F, t=t, timedep_coeffs=(u,))
+        F_dtless = remove_time_derivatives(split_form.time)
         F_remainder = expand_time_derivatives(split_form.remainder, t=t, timedep_coeffs=())
-        u_np = to_value(self.u0, self.stages, self.vandermonde)
+
+        Fupdate = replace(F_dtless, {u: unew}) - replace(F_dtless, {u: u0})
+
+        C = vecconst(self.butcher_tableau.c)
+        B = vecconst(self.butcher_tableau.b)
```
pbrubeck:
```suggestion
```

irksome/stage_value.py
```diff
@@ -220,22 +233,49 @@ def _update_stiff_acc(self):
...
+        weighted mass matrix rather than the stage operator.  Stage-
+        tuned options such as fieldsplit indices, ``snes_type='ksponly'``,
+        lagged Jacobians, or custom multigrid transfers generally do
+        not apply.  If ``update_solver_parameters`` is None, Firedrake's
+        default solver parameters are used (typically a sparse direct
+        solve).  Pass an explicit dict to override.
+        """
         # only form update stuff if we need it
         # which means neither stiffly accurate nor Vandermonde
         backend_cls = self._backend
```
pbrubeck, line 266:
```suggestion
        backend_cls = self._backend
        C = vecconst(self.butcher_tableau.c)
        B = vecconst(self.butcher_tableau.b)
```

irksome/stage_value.py
sghelichkhani: Yes, `g = c·id` is fine, and more generally any `Dt(L(u_0))` with `L` linear in `u_0` flows through `_update_Ainv` correctly. The comment is actually then narrower than the contrac. The classifier doesn't test for literal identity. It takes the Gateaux derivative of the `Dt` operand and asks whether `u_0` still appears. So for `Dt(c*u)`, `Dt(c(x)*u)`, `Dt(M*u)` with `M` a matrix constant, or `Dt(u + g(t,x))`, the derivative drops `u_0` entirely, so they're classified as linear and routed to bAinv. And the bAinv reconstruction is exactly conservative for any linear `L`: the stage equations enforce `inner(v, L(U_i) - L(u_0)) = -dt\cdot \sum_j A_ij F_rem(U_j)`, and multiplying by `bAinv_i`, summing, and using linearity of `L` together with `A^T·bAinv = b` gives back `inner(v, L(u_new) - L(u_0)) = -dt\cdot \sum_j b_j F_rem(U_j)`. So no `g = id` special-casing involved it's the linearity of `L` that does the work.

There's also a second reason to keep `_update_Ainv` on the linear branch that the comment doesn't mention: DAE structure. For forms with an algebraic block (no `Dt` on some components), the conservative variational head has a singular Jacobian on that block, whereas the bAinv linear combination propagates the algebraic stage values into `u_new` correctly. So bAinv isn't just a fast path for the linear case, to me it is the only correct one when there's a DAE component (correct me if wrong!)

I'll tighten the comment to say "linear in `u_0`" and add the DAE rationale, and pin both halves with tests: 
    - a classifier test that asserts `has_nonlinear_time_derivative` returns `False` for `Dt(c*u)`, `Dt(Constant(c)*u)`, `Dt(c(x)*u)`, and `Dt(u + sin(t))`
    - a numerical mass-conservation test with a scaled mass term `inner(v, Dt(c*u))*dx` on a non-SA tableau, asserting machine precision.

Will push next.

rckirby (approved):
Thanks for the contribution!

PR #227, opened 2026-05-14
Tidy up TrialFunction/LinearVariationalSolver support

pbrubeck:
Improvements upon #213

Avoids construction of an intermediate semi discrete linear form and goes straight from semi discrete bilinear form into the stage bilinear form.

Consistently use LVP/LVS across all time stepper classes.

### DIFF
--- a/irksome/dirk_stepper.py
+++ b/irksome/dirk_stepper.py
@@ -1,18 +1,15 @@
 import numpy
-from firedrake import (derivative, Function,
-                       LinearVariationalSolver,
-                       NonlinearVariationalProblem,
-                       NonlinearVariationalSolver)
-from ufl.constantvalue import as_ufl
+from ufl import as_ufl, lhs
 
-from .ufl.deriv import TimeDerivative, expand_time_derivatives
 from .constant import vecconst
-from .tools import replace
-from .constant import MeshConstant
+from .backend import get_backend
 from .bcs import bc2space
+from .ufl.deriv import TimeDerivative, expand_time_derivatives
+from .tools import replace
 
 
-def getFormDIRK(F, ks, butch, t, dt, u0, bcs=None, kgac=None):
+def getFormDIRK(F, ks, butch, t, dt, u0, bcs=None, kgac=None, backend="firedrake"):
+    backend_cls = get_backend(backend)
     if bcs is None:
         bcs = []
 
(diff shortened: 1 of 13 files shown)


irksome/dirk_stepper.py
```diff
@@ -132,9 +133,10 @@ def __init__(self, F, butcher_tableau, t, dt, u0, bcs=None, Fp=None,
 
         stage_Jp = None
         if Fp is not None:
-            Fp = as_linear_form(Fp, u0)
-            stage_Fp, *_ = self.get_form_and_bcs(self.ks, F=Fp, bcs=())
-            stage_Jp = derivative(stage_Fp, k)
+            Fp_linear = len(Fp.arguments()) == 2
+            ks_Fp = Fp.arguments()[1] if Fp_linear else self.ks
+            stage_Fp, *_ = self.get_form_and_bcs(ks_Fp, F=Fp, bcs=())
+            stage_Jp = stage_Fp if Fp_linear else derivative(stage_Fp, k)
```
```suggestion
            stage_Jp = lhs(stage_Fp) if Fp_linear else derivative(stage_Fp, k)
```

irksome/dirk_stepper.py
```suggestion
from ufl import as_ufl, lhs
```

irksome/imex.py
```diff
@@ -41,39 +38,43 @@ def getFormExplicit(Fexp, butch, u0, UU, t, dt, splitting=None):
...
     Aprop = vecconst(Aexp)
     Ait = vecconst(butch.A)
     C = vecconst(butch.c)
 
     v_np = reshape(VV, (num_stages, *u0.ufl_shape))
     u_np = reshape(UU, (num_stages, *u0.ufl_shape))
 
-    Fit = zero()
-    Fprop = zero()
+    Fit = Form([])
```
jorgensd, line 59: You could potentially use `ZeroBaseForm` here: https://github.com/FEniCS/ufl/blob/1676aec1dc1b92e4f9b2bb007659bb45d6a007f7/ufl/form.py#L863-L868 which you could give `VV` as input, i.e. `Fprop = ZeroBaseForm((VV,))`.
Not sure it is supported in firedrake, but it would give you a form with the right arity irregardless of If anything is added to the form (as far as I can tell).

irksome/imex.py
```diff
@@ -308,29 +310,39 @@ def invalidate_jacobian(self):
...
-    # preprocess time derivatives
-    F = expand_time_derivatives(F, t=t, timedep_coeffs=(u0,))
-    Fexp = expand_time_derivatives(Fexp, t=t, timedep_coeffs=(u0,))
-
-    v = F.arguments()[0]
+    try:
+        v, u = F.arguments()
+    except ValueError:
+        v, = F.arguments()
+        u = u0
```
jorgensd: As this is repeated in quite a few places, should we have a convenience function that is `v, u = extract_arguments(F, u0)` ?

pbrubeck: I thought about that, we definitely should. But we should come up with a better name (that function already exists in ufl and does a different thing)

irksome/imex.py
pbrubeck: We support ZeroBaseForm (and also introduced it). I don't think it's any better. As we need extra logic for linear vs bilinear Form

jorgensd (approved):

PR #231, opened 2026-05-18
Make release.yml look like mpi-pytest

rckirby:
This implements @connorjward 's suggestion for the right workflow to make releases on PyPi


connorjward (approved):
LGTM

PR #233, opened 2026-05-20
Submesh demo

pbrubeck:


rckirby (changes requested):
Nice-looking demo, but fails to run.  Do we just need to pip install netgen/ngspetsc in our CI?

PR #245, opened 2026-06-08
Support quasi-Newton methods

pbrubeck:


irksome/dirk_stepper.py
```diff
@@ -69,7 +69,8 @@ class DIRKTimeStepper:
     """Front-end class for advancing a time-dependent PDE via a diagonally-implicit
     Runge-Kutta method formulated in terms of stage derivatives."""
 
-    def __init__(self, F, butcher_tableau, t, dt, u0, bcs=None, Fp=None,
+    def __init__(self, F, butcher_tableau, t, dt, u0, bcs=None,
+                 J=None, Jp=None, scheme_J=None, scheme_Jp=None,
```
rckirby: Not sure we want to allow re-discretization in a DIRK. Is there anything to be gained by it when we already solve stage-by-stage?

irksome/multistep.py
```diff
@@ -38,8 +38,10 @@ class MultistepTimeStepper(BaseTimeStepper):
             to find the required starting values.
     """
 
-    def __init__(self, F, method, t, dt, u0, bcs=None, Fp=None, solver_parameters=None, bounds=None, appctx=None, nullspace=None,
-                 transpose_nullspace=None, near_nullspace=None, startup_parameters=None, backend: str = "firedrake", **kwargs):
+    def __init__(self, F, method, t, dt, u0, bcs=None, J=None, Jp=None, solver_parameters=None, bounds=None, appctx=None, nullspace=None,
```
rckirby, line 40: Also, no point in rediscretising multistep methods with other multistep methods in time, it won't simplify the system.

irksome/nystrom_dirk_stepper.py
```diff
@@ -98,7 +98,7 @@ class DIRKNystromTimeStepper:
     """Front-end class for advancing a second-order time-dependent PDE via a diagonally-implicit
     Runge-Kutta-Nystrom method formulated in terms of stage derivatives."""
 
-    def __init__(self, F, tableau, t, dt, u0, ut0, bcs=None, Fp=None,
+    def __init__(self, F, tableau, t, dt, u0, ut0, bcs=None, J=None, Jp=None,
```
rckirby, line 100: Ditto for Dirk Nystrom.

irksome/stepper.py
```diff
@@ -11,29 +11,29 @@
...
+valid_base_kwargs = ("bcs", "J", "Jp", "form_compiler_parameters",
                      "is_linear", "constant_jacobian",
                      "restrict", "solver_parameters",
                      "nullspace", "transpose_nullspace", "near_nullspace",
                      "appctx", "options_prefix", "pre_apply_bcs")
 
 valid_kwargs_per_stage_type = {
-    "deriv": ["Fp", "stage_type", "bc_type", "splitting", "adaptive_parameters", "aux_indices", "sample_points"],
-    "value": ["Fp", "stage_type", "basis_type",
+    "deriv": ["stage_type", "scheme_J", "scheme_Jp", "bc_type", "splitting", "adaptive_parameters", "aux_indices", "sample_points"],
```
rckirby, line 21: Do we have need/use case for rediscretising IRK with other IRK?  I guess one could implement Rana-like things this way if one really wanted to.  But we've already got other avenues for that.

irksome/stepper.py
```diff
@@ -11,29 +11,29 @@
...
+    "value": ["stage_type", "scheme_J", "scheme_Jp", "basis_type",
               "update_solver_parameters", "splitting", "bounds", "use_collocation_update", "sample_points"],
-    "dirk": ["Fp", "stage_type"],
-    "explicit": ["Fp", "stage_type"],
+    "dirk": ["stage_type", "scheme_J", "scheme_Jp"],
```
rckirby: Again, DIRK/explicit probably don't need these?

irksome/stepper.py
```diff
@@ -188,28 +193,30 @@ def TimeStepper(F, method, t, dt, u0, **kwargs):
...
             update_solver_parameters=update_solver_parameters,
             bounds=bounds, use_collocation_update=use_collocation_update,
             sample_points=sample_points,
             **base_kwargs)
     elif stage_type == "dirk":
-        Fp = kwargs.get("Fp", None)
         return DIRKTimeStepper(
-            F, method, t, dt, u0, bcs, Fp=Fp, **base_kwargs)
+            F, method, t, dt, u0, bcs,
+            scheme_J=scheme_J, scheme_Jp=scheme_Jp,
```
rckirby: Here too if we're deleting.

irksome/stepper.py
```diff
@@ -188,28 +193,30 @@ def TimeStepper(F, method, t, dt, u0, **kwargs):
...
+            **base_kwargs)
     elif stage_type == "explicit":
-        Fp = kwargs.get("Fp", None)
         return ExplicitTimeStepper(
-            F, method, t, dt, u0, bcs, Fp=Fp, **base_kwargs)
+            F, method, t, dt, u0, bcs,
+            scheme_J=scheme_J, scheme_Jp=scheme_Jp,
```
rckirby: Here

rckirby (commented):
Thanks, looks good, but I think we don't want to introduce rediscretization for dirk/explicit/etc.  Maybe we only want this for GiT, but FIRK is worth considering.

irksome/dirk_stepper.py
pbrubeck: I propagated `J, Jp, scheme_J, scheme_Jp` to the implicit steppers. Maybe it does not make much sense to change the scheme for stage-decoupled steppers. But my goal was to keep things more or less uniform to make them easier to mantain.

irksome/stepper.py
pbrubeck: We could do Rana through the front-door.

irksome/multistep.py
```diff
@@ -38,8 +38,10 @@ class MultistepTimeStepper(BaseTimeStepper):
...
+                 transpose_nullspace=None, near_nullspace=None, startup_parameters=None, backend: str = "firedrake",
+                 scheme_J=None, scheme_Jp=None,
```
pbrubeck:
```suggestion
```

irksome/multistep.py
```diff
@@ -51,15 +53,11 @@ def __init__(self, F, method, t, dt, u0, bcs=None, Fp=None, solver_parameters=No
...
         self.us.append(u0)
         Fnew, bcsnew = self.get_form_and_bcs(F, t, dt, u0, self.a, self.b, bcs=bcs)
-
-        if Fp is not None:
-            Fpnew, _ = self.get_form_and_bcs(Fp, t, dt, u0, self.a, self.b, bcs=bcs)
-            Jp = self._backend.derivative(Fpnew, self.us[-1])
-        else:
-            Jp = None
+        J = self.get_bilinear_form(J, u0, method=scheme_J)
+        Jp = self.get_bilinear_form(Jp, u0, method=scheme_Jp)
```
pbrubeck:
```suggestion
        J = self.get_bilinear_form(J, u0)
        Jp = self.get_bilinear_form(Jp, u0)
```

irksome/multistep.py
```diff
@@ -162,6 +162,16 @@ def get_form_and_bcs(self, F, t, dt, u0, a, b, bcs=None):
 
         return Fnew, bcsnew
 
+    def get_bilinear_form(self, form, u0, method=None):
+        if method is not None:
+            raise NotImplementedError("Cannot change the method for the preconditioner")
```
pbrubeck:
```suggestion
    def get_bilinear_form(self, form, u0):
```

irksome/nystrom_dirk_stepper.py
pbrubeck: I would leave `J` and `Jp`, it still makes sense to do quasi-Newton or use a different form for preconditioning (e.g. for Augmented Lagrangian)

irksome/multistep.py
pbrubeck: I deleted `scheme_J` and `scheme_Jp` from here

irksome/stepper.py
```diff
@@ -11,29 +11,29 @@
...
 valid_kwargs_per_stage_type = {
-    "deriv": ["Fp", "stage_type", "bc_type", "splitting", "adaptive_parameters", "aux_indices", "sample_points"],
-    "value": ["Fp", "stage_type", "basis_type",
+    "deriv": ["stage_type", "scheme_J", "scheme_Jp", "bc_type", "splitting", "adaptive_parameters", "aux_indices", "sample_points"],
+    "value": ["stage_type", "scheme_J", "scheme_Jp", "basis_type",
               "update_solver_parameters", "splitting", "bounds", "use_collocation_update", "sample_points"],
-    "dirk": ["Fp", "stage_type"],
-    "explicit": ["Fp", "stage_type"],
+    "dirk": ["stage_type", "scheme_J", "scheme_Jp"],
+    "explicit": ["stage_type", "scheme_J", "scheme_Jp"],
```
pbrubeck:
```suggestion
    "dirk": ["stage_type"],
    "explicit": ["stage_type"],
```

irksome/stepper.py
```diff
@@ -188,28 +193,30 @@ def TimeStepper(F, method, t, dt, u0, **kwargs):
...
+            F, method, t, dt, u0, bcs,
+            scheme_J=scheme_J, scheme_Jp=scheme_Jp,
+            **base_kwargs)
     elif stage_type == "explicit":
-        Fp = kwargs.get("Fp", None)
         return ExplicitTimeStepper(
-            F, method, t, dt, u0, bcs, Fp=Fp, **base_kwargs)
+            F, method, t, dt, u0, bcs,
+            scheme_J=scheme_J, scheme_Jp=scheme_Jp,
+            **base_kwargs)
```
pbrubeck:
```suggestion
            F, method, t, dt, u0, bcs, **base_kwargs)
```

irksome/stepper.py
```diff
@@ -188,28 +193,30 @@ def TimeStepper(F, method, t, dt, u0, **kwargs):
...
             bounds=bounds, use_collocation_update=use_collocation_update,
             sample_points=sample_points,
             **base_kwargs)
     elif stage_type == "dirk":
-        Fp = kwargs.get("Fp", None)
         return DIRKTimeStepper(
-            F, method, t, dt, u0, bcs, Fp=Fp, **base_kwargs)
+            F, method, t, dt, u0, bcs,
+            scheme_J=scheme_J, scheme_Jp=scheme_Jp,
+            **base_kwargs)
```
pbrubeck:
```suggestion
            F, method, t, dt, u0, bcs, **base_kwargs)
```

irksome/dirk_stepper.py
```suggestion
    def __init__(self, F, butcher_tableau, t, dt, u0, bcs=None, J=None, Jp=None,
```

irksome/dirk_stepper.py
```diff
@@ -127,11 +128,10 @@ def __init__(self, F, butcher_tableau, t, dt, u0, bcs=None, Fp=None,
...
 
-        stage_Jp = None
-        if Fp is not None:
-            Fp_linear = len(Fp.arguments()) == 2
-            stage_Fp, *_ = self.get_form_and_bcs(self.ks, F=Fp, bcs=())
-            stage_Jp = lhs(stage_Fp) if Fp_linear else backend_cls.derivative(stage_Fp, k)
+        scheme_J = scheme_J or butcher_tableau
+        scheme_Jp = scheme_Jp or scheme_J
+        stage_J = self.get_bilinear_form(J, u0, k, tableau=scheme_J)
+        stage_Jp = self.get_bilinear_form(Jp, u0, k, tableau=scheme_Jp)
```
```suggestion
        stage_J = self.get_bilinear_form(J, u0, k, tableau=butcher_tableau)
        stage_Jp = self.get_bilinear_form(Jp, u0, k, tableau=butcher_tableau)
```

irksome/dirk_stepper.py
rckirby: It could make sense to access alternate schemes this way for IRK, such as quasi-Newton where you Rana the Jacobian (instead of in the preconditioner).  

Uniformity is easier to maintain, but if we make it easy for users to do nonsensical things, then we have to figure out whether to debug the maths or our code ;)

irksome/multistep.py
rckirby: I think this is wise.  There is simply nothing to be gained on the solver front from approximating the multistep Jacobian with the Jacobian from a different multistep one -- they're both morally scaled backward Euler.

irksome/stepper.py
rckirby: Quasi-Newton -- if you wanted to introduce the approximation at the Newton rather than Krylov level.  That's the only kind of situation where I could imagine one would want to approximate an IRK method with an IRK method.  But maybe there are others and we can leave this.

irksome/galerkin_stepper.py
```diff
@@ -109,6 +124,7 @@ def getTermGalerkin(F, L_trial, L_test, Q, t, dt, u0, stages, test, aux_indices,
                    v: vsub[q] * dt,
                    u: usub[q],
                    dtu: dtusub[q] / dt}
+
```
```suggestion
```

rckirby (approved):

PR #246, opened 2026-06-11
Raise error for invalid `stage_type`

JHopeCollins:
Otherwise if you are silly and pass something like `stage_type="stage_deriv"` to the `TimeStepper` function then it just silently returns `None` instead of an actual object.


rckirby (approved):
Good catch, this looks reasonable to me!

PR #228, opened 2026-05-15
Lag label

danshapero:
Resolves #221. This PR adds a function `lag` which labels a single expression in a form for holding back to the start of a timestep. As @pbrubeck thought, implementing it amounted to modifying `irksome.tools.replace`. I added a test based on lagging the conductivity in the Stefan problem. Without lagging, that fails because of the conditional.

In addition to giving you an escape hatch, you can use lagging to define nonlinear preconditioners. I have a demo of this for the Allen-Cahn equation coded up just using petsc4py which I think is pretty compelling. @JHopeCollins and I are working on getting AuxiliaryOperatorSNES merged into Firedrake: firedrakeproject/firedrake#5119. Once that's merged I can write an Irksome demo showing that off, either in this PR or a separate one.

PR #156, opened 2025-06-20
Stage Functions

pbrubeck:

PR #248, opened 2026-06-19
Nonlinear preconditioning

danshapero:
Firedrake can use PETSc's nonlinear preconditioning features through the `AuxiliaryOperatorSNES` class. This patch adds a class `IRKAuxiliarySNES` which is almost identical to `IRKAuxiliaryOperatorPC` so that Irksome can use nonlinear preconditioning.

I added a test case based on the Stefan problem. This is similar to the test case for lagging that I used in #228. But now the lagged form is a preconditioner to the fully implicit form.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
   test:
     runs-on: [self-hosted, Linux]
     container:
-      image: firedrakeproject/firedrake-vanilla-default:latest
+      image: firedrakeproject/firedrake-vanilla-default:dev-release
     steps:
       - name: Fix HOME
         # For unknown reasons GitHub actions overwrite HOME to /github/home
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,9 +15,12 @@ concurrency:
 
 jobs:
   test:
+    strategy:
+      matrix:
+        version: [2026.4.0, dev-release]
     runs-on: [self-hosted, Linux]
     container:
-      image: firedrakeproject/firedrake-vanilla-default:dev-release
+      image: firedrakeproject/firedrake-vanilla-default:${{ matrix.version }}
     steps:
       - name: Fix HOME
         # For unknown reasons GitHub actions overwrite HOME to /github/home
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
   test:
     strategy:
       matrix:
-        version: [2026.4.0, dev-release]
+        version: [latest, dev-release]
     runs-on: [self-hosted, Linux]
     container:
       image: firedrakeproject/firedrake-vanilla-default:${{ matrix.version }}
--- a/irksome/pc.py
+++ b/irksome/pc.py
@@ -82,13 +82,13 @@ class IRKAuxiliaryOperatorPC(AuxiliaryOperatorPC):
 
 
 class IRKAuxiliaryOperatorSNES(AuxiliaryOperatorSNES):
-    """Base class that inherits from Firedrake's :class:`AuxiliaryOperatorSNES`
+    """Base class that inherits from Firedrake's auxiliary operator SNES
     and provides the nonlinear form associated with an auxiliary Form and/or
     approximate Butcher matrix (which are provided by subclasses). This is the
     nonlinear analogue of :class:`IRKAuxiliaryOperatorPC`.
 
     Options for the inner solve are specified using the ``"aux_"`` prefix, as
-    for :class:`AuxiliaryOperatorSNES`.
+    for the base auxiliary operator SNES.
     """
 
     def getNewForm(self, snes, u0, test):
--- a/irksome/pc.py
+++ b/irksome/pc.py
@@ -2,7 +2,7 @@ import copy
 
 import numpy
 from .labeling import as_form
-from firedrake import AuxiliaryOperatorPC, derivative
+from firedrake import AuxiliaryOperatorPC, AuxiliaryOperatorSNES, derivative
 from firedrake.dmhooks import get_appctx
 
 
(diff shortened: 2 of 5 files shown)


tests/test_npc.py
```diff
@@ -0,0 +1,107 @@
...
+    }
+    method = BackwardEuler()
+    stepper = TimeStepper(F, method, t, dt, T, **params)
+
+    final_time = 20.0
+    num_steps = int(final_time / float(dt))
+    for _ in range(num_steps):
+        stepper.advance()
+        t.assign(t + dt)
+        assert stepper.solver.snes.getConvergedReason() > 0
```
pbrubeck: We should assert that we get a reasonably accurate solution or conservation/dissipation of some quantity

danshapero: I changed this to assert that it converged due to FNORM_RELATIVE. The boundary conditions are oscillating, so unlike the case of running to steady state there isn't an obvious dissipation potential. If I change it to run to steady state then you get into annoying issues of setting an absolute tolerance. If you think it's better to use this as a demo than a unit test then I can think of something else to try instead. I just wanted something that will obviously break Newton.

rckirby: I think breaking/not breaking Newton as expected as a test is fine.  But, since this is a new feature, if you're offering to write a demo, we'd gladly take it.  It could be a separate PR, and might take some text explaining what the nonlinear PC is doing, etc.

rckirby (approved):
I'm happy to approve this.  I'd like to eventually have a demo, but that could be a separate, later contribution.

irksome/pc.py
```diff
@@ -81,6 +81,65 @@ def form(self, pc, test, trial):
         return Jnew, bcnew
 
 
+class IRKAuxiliaryOperatorSNES(AuxiliaryOperatorSNES):
+    """Base class that inherits from Firedrake's :class:`AuxiliaryOperatorSNES`
```
JHopeCollins: @pbrubeck will this resolve to link to the Firedrake docs? Does Irksome have the intersphinx set up?
Even with intersphinx it might have to be `:class:`~firedrake.AuxiliaryOperatorSNES`` to resolve correctly.

irksome/pc.py
```diff
@@ -81,6 +81,65 @@ def form(self, pc, test, trial):
...
+        """Derived classes produce a typically structured
+        approximation to A."""
+        raise NotImplementedError
+
+    def form(self, snes, w0, w, test):
+        """Implements the interface for AuxiliaryOperatorSNES.
+
+        :arg snes: the PETSc SNES object.
+        :arg w0: the current iterate of the stages (unused by default, but
+            available to subclasses that wish to lag terms).
```
JHopeCollins, line 161: @pbrubeck whats the state of https://github.com/firedrakeproject/Irksome/pull/156? We'd need that functionality to properly use `w0` here.

If that PR isn't ready soon then we can at least raise an issue to fix this SNES once that PR is merged.

tests/test_npc.py
```diff
@@ -0,0 +1,114 @@
...
+            },
+        },
+    }
+    method = BackwardEuler()
+    stepper = TimeStepper(F, method, t, dt, T, **params)
+
+    reasons = [
+        PETSc.SNES.ConvergedReason.CONVERGED_FNORM_ABS,
+        PETSc.SNES.ConvergedReason.CONVERGED_FNORM_RELATIVE,
+        PETSc.SNES.ConvergedReason.CONVERGED_SNORM_RELATIVE,
```
JHopeCollins: Are we ok with stagnation here? We also aren't setting the `snes_atol` in the options dictionary so testing that convergence isn't robust to future changes.
If you set `snes_stol: 0` and an explicit `snes_atol` (possibly also 0) then you can just check the solver didn't diverge.

irksome/pc.py
danshapero: The only way that I can get this to work is with the fully-qualified name:
```
:class:`~firedrake.preconditioners.auxiliary_snes.AuxiliaryOperatorSNES`
```
But Sphinx automatically generates a correct link to AuxiliaryOperatorSNES right above the docstring itself because it's a base class. It seems redundant to have two links to the same class all within a paragraph of each other in the docstring, so I changes those links to regular text.

tests/test_npc.py
danshapero: I tried `snes_stol: 0` and this test looks a bit worse now. Most of the time it converges due to fnorm_relative. Every so often the solver will stagnate but then go back to normal at the next timestep. I wasn't able to get rid of the stagnation by messing with solver options. I get the same behavior even if I do the timestepping manually and use the previous guess of the solver (i.e. true Picard iteration) as a NPC. I can work up another test case using Allen-Cahn or something we already know if we're not ok with occasional stagnation.

JHopeCollins: Personally I would prefer not equating stagnating with converging, so if there is an alternative I'd prefer that. But I am not familiar at all with this equation so maybe there is a detail here I'm missing. If so then it'd be useful to have a quick comment next to the `reasons` to explain why we're ok with snorm sometimes.

irksome/pc.py
pbrubeck: The PR is stale. I remember things were in a good state when I last worked on it.

tests/test_npc.py
danshapero: I changed the test case to use a very sharp but smooth transition between the solid and liquid phases. It breaks Newton but works with the lagged NPC.

.github/workflows/ci.yml
```diff
@@ -15,9 +15,12 @@ concurrency:
 
 jobs:
   test:
+    strategy:
+      matrix:
+        version: [2026.4.0, dev-release]
```
JHopeCollins: Saves having to update this number each time there's a new release.
```suggestion
        version: [latest, dev-release]
```

PR #250, opened 2026-06-30
Fix IRKAuxiliaryOperatorPC + Jp

pbrubeck:


irksome/galerkin_stepper.py
```diff
@@ -87,7 +87,7 @@ def getTermGalerkin(F, L_trial, L_test, Q, t, dt, u0, stages, test, aux_indices,
 
     qpts = Q.get_points()
     qwts = Q.get_weights()
-    #assert qpts.size >= L_test.space_dimension()
+    # assert qpts.size >= L_test.space_dimension()
```
rckirby: Since this is commented out, can we just delete it before merging?

rckirby (changes requested):
Two minor requests.  One inline below.  The other is that there seems to be a mismatch between the title of the PR and what changes you're making.  Can you please update the title and/or add a clearer description what problem you're solving and why these changes solve it?

irksome/galerkin_stepper.py
pbrubeck: Sure. I commented this out because this assertion affects assembly of lower order RHS terms

irksome/galerkin_stepper.py
```diff
@@ -337,7 +336,7 @@ def get_form_and_bcs(self, stages, F=None, bcs=None,
             order = self.order
         aux_indices = aux_indices or self.aux_indices
 
-        if isinstance(tableau, CollocationButcherTableau):
+        if isinstance(tableau, ButcherTableau):
```
rckirby, line 339: The Git <-> IRK connection via quadrature only makes sense for collocation schemes.  Is there a compelling reason to allow something more general here?

rckirby (changes requested):
Also, please add a description of what the PR is doing!

irksome/galerkin_stepper.py
pbrubeck: Yes, there is. I introduced support for Jp=RanaLDScheme, which inherits from ButcherTableau. Should it inherit from ButcherCollocation?

rckirby: Oh, I had figured users would just want to continue getting Rana from a Python PC.  If we want to allow Rana via Jp (which might be reasonable), maybe we want to insert some base class for Rana-like things and check that we're a collocation tableau or one of those.  After all, Rana isn't really collocation, it's some kind of DIRK we don't have any theory for.

tests/test_pc.py
```diff
@@ -81,17 +78,23 @@ def rd(scheme, **kwargs):
...
+        basis_type = scheme.basis_type
+        if isinstance(basis_type, tuple):
+            basis_type = basis_type[1]
+        if basis_type == "radau":
+            return RadauIIA(scheme.num_stages)
+        else:
+            raise ValueError(f"Unexpected basis_type {basis_type}")
+
+    butcher = scheme if isinstance(scheme, ButcherTableau) else to_butcher(scheme)
+    schemeDU = RanaDUScheme(butcher)
```
pbrubeck:
```suggestion
    schemeDU = RanaDUScheme(to_butcher(scheme))
```

tests/test_pc.py
```diff
@@ -81,17 +78,23 @@ def rd(scheme, **kwargs):
         }
     }
 
+    def to_butcher(scheme):
+        basis_type = scheme.basis_type
```
pbrubeck:
```suggestion
    def to_butcher(scheme):
        if isinstance(scheme, ButcherTableau):
            return scheme 
        basis_type = scheme.basis_type
```

irksome/galerkin_stepper.py
```suggestion
            raise TypeError("Expecting ButcherTableau or ContinuousPetrovGalerkinScheme")
```

irksome/pc.py
```diff
@@ -100,6 +104,34 @@ def getAtilde(self, A):
...
+    if isinstance(scheme, ButcherTableau):
+        return scheme
+
+    basis_type = scheme.basis_type
+    order = scheme.num_stages
+    if isinstance(basis_type, tuple):
+        basis_type = basis_type[1]
+    if basis_type == "radau":
+        return RadauIIA(order)
+    else:
```
rckirby: Why not Gauss-Legendre here as well?

pbrubeck: It's supposed to be handled by the generic case... I think this helper needs testing.

pbrubeck: I removed special casing for RadauIIA.

PR #251, opened 2026-07-09
DAE support in StageDerivativeTimeStepper

pbrubeck:


irksome/stage_derivative.py
```diff
@@ -201,6 +204,16 @@ def __init__(self, F, butcher_tableau, t, dt, u0, bcs=None,
             raise NotImplementedError("A=A1 A2 splitting needs A2 invertible")
 
         self.aux_indices = aux_indices
+        self.aux_updateb = None
+        self.aux_update_scale = None
+        if aux_indices:
+            try:
+                stage_update = numpy.linalg.solve(butcher_tableau.A.T, butcher_tableau.b)
+                self.aux_updateb = vecconst(numpy.linalg.solve(A2.T, stage_update), backend=backend)
+                self.aux_update_scale = 1 - numpy.sum(stage_update)
```
pbrubeck, line 213: I didn't have to do the maths, codex figured out that the correct linear combination for the update should match that of the stage-value formulation.

irksome/stage_derivative.py
```diff
@@ -201,6 +204,16 @@ def __init__(self, F, butcher_tableau, t, dt, u0, bcs=None,
             raise NotImplementedError("A=A1 A2 splitting needs A2 invertible")
 
         self.aux_indices = aux_indices
+        self.aux_updateb = None
+        self.aux_update_scale = None
+        if aux_indices:
+            try:
+                stage_update = numpy.linalg.solve(butcher_tableau.A.T, butcher_tableau.b)
+                self.aux_updateb = vecconst(numpy.linalg.solve(A2.T, stage_update), backend=backend)
```
pbrubeck, line 212: However, given that A = A1 * A2, I think this could be simplified into a single linear system solve involving A1
```suggestion
                stage_update = numpy.linalg.solve(butcher_tableau.A1.T, butcher_tableau.b)
                self.aux_updateb = vecconst(stage_update, backend=backend)
```
Something along those lines

irksome/stepper.py
```diff
@@ -111,9 +111,9 @@ def TimeStepper(F, method, t, dt, u0, **kwargs):
         enforcing bounds constraints with an RK method that is not stiffly
         accurate. Currently, only constant-in-time boundary conditions are
         supported.
-    :kwarg aux_indices: Only valid for continuous Petrov Galerkin time scheme.  It
-        specifies that some of the variables in `u0` are to be treated as
-        auxiliary, that is, discretized in the lower-order DG test space.
+    :kwarg aux_indices: Valid for continuous Petrov Galerkin schemes and
+        derivative-stage RK schemes. It specifies that some variables in `u0`
```
pbrubeck, line 115:
```suggestion
    :kwarg aux_indices: Valid for stage-derivative RK and continuous Petrov
        Galerkin schemes. It specifies that some variables in `u0`
```

irksome/stage_derivative.py
```diff
@@ -222,7 +235,12 @@ def _update(self):
 
         # Note: this now catches the optimized/stiffly accurate case as b[s] == Zero() will get dropped
         for i, u0bit in enumerate(self.u0.subfunctions):
-            u0bit += sum(self.stages.subfunctions[nf * s + i] * (b[s] * dt) for s in range(ns))
+            if self.aux_indices and i in self.aux_indices:
+                u0bit *= self.aux_update_scale
+                u0bit += sum(self.stages.subfunctions[nf * s + i] * (self.aux_updateb[s] * dt)
+                             for s in range(ns))
+            else:
+                u0bit += sum(self.stages.subfunctions[nf * s + i] * (b[s] * dt) for s in range(ns))
```
ScottMacLachlan, line 243: This introduces an `if` statement into the general `_update` formula which is only needed for an edge case.  I would argue this deserves a special update function that is selected at `__init__` time and not foisted on every other use case

ScottMacLachlan (changes requested):
This does not make sense to me.

irksome/stage_derivative.py
pbrubeck: It is what we have in Galerkin, happy to split it into its own function

rckirby: It's probably a minor ding on run-time given the cost of solving systems, but if we can lift `if` statements out of `do` loops, I'm in favor.

irksome/stage_derivative.py
rckirby: I recall there being problems with updates in the non-stiffly accurate case.  I suspect this may still fail silently and compute the wrong answer if you have things not correctly labeled as auxiliary.  So some tests might be in order.

irksome/stage_derivative.py
rckirby: Please confirm and get that correct?

rckirby (changes requested):
See comments already made.  Also, please put an actual description of what problem you're fixing/feature you're adding in the PR, as it will make getting started on reviewing much easier!

